PR #3, opened 2020-04-28
Add entityframework objects to replace manual SQL lite.

Chrafangau:
To use, add a public constructor that takes a FloofDataContext object. DB is called like this:
_floofDB.Add(new Tag { Content = "testTag1", TagId = 1, UserId = Context.User.Id});
_floofDB.SaveChanges();
_floofDB.Tags.Find(1UL);


Floofbot/Migrations/FloofDataContextModelSnapshot.cs
```diff
@@ -0,0 +1,113 @@
...
+                    b.Property<string>("Reason")
+                        .HasColumnType("TEXT");
+
+                    b.Property<ulong>("UserId")
+                        .HasColumnType("INTEGER");
+
+                    b.HasKey("Id");
+
+                    b.ToTable("Warnings");
+                });
```
iShauny, line 109: Is this a different schema from nadeko? iirc it was to remain the same so that warnings can be moved over easily

PR #5, opened 2020-04-29
Updated Event Logging to Work With Entity Framework

iShauny:
The structure of the module has been updated to work with Entity Framework instead of directly Sqlite. Also, some basic exception handling has been added.

### DIFF
--- a/Floofbot/Modules/Logging.cs
+++ b/Floofbot/Modules/Logging.cs
@@ -191,13 +191,7 @@ namespace Floofbot.Modules
                 if (ServerConfig == null) // no entry in DB for server - not configured
                     return false;
 
-                bool bEnabled = ServerConfig.IsOn;
-                if (!bEnabled)
-                    return false;
-                else if (bEnabled)
-                    return true;
-                else
-                    return false;
+                return ServerConfig.IsOn;
             }
 
             public async Task MessageUpdated(Cacheable<IMessage, ulong> before, SocketMessage after, ISocketMessageChannel chan)
--- a/Floofbot/Modules/Logging.cs
+++ b/Floofbot/Modules/Logging.cs
@@ -482,7 +482,6 @@ namespace Floofbot.Modules
                 catch (Exception ex)
                 {
                     Log.Error("Error with the guild member updated event handler: " + ex);
-                    Console.Write(ex);
                     return;
                 }
 
--- a/Floofbot/Modules/Logging.cs
+++ b/Floofbot/Modules/Logging.cs
@@ -429,6 +429,9 @@ namespace Floofbot.Modules
                             .AddField("New Name", user.Username)
                             .WithFooter(DateTime.Now.ToString());
 
+                        if (Uri.IsWellFormedUriString(user.GetAvatarUrl(), UriKind.Absolute))
+                            embed.WithThumbnailUrl(user.GetAvatarUrl());
+
                     }
                     else if (before.Nickname != after.Nickname)
                     {
@@ -438,6 +441,9 @@ namespace Floofbot.Modules
                             .AddField("Old Nickname", before.Nickname)
                             .AddField("New Nickname", user.Nickname)
                             .WithFooter(DateTime.Now.ToString());
+
+                        if (Uri.IsWellFormedUriString(user.GetAvatarUrl(), UriKind.Absolute))
+                            embed.WithThumbnailUrl(user.GetAvatarUrl());
 
                     }
                     else if (before.AvatarId != after.AvatarId)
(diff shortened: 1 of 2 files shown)


Floofbot/Modules/Logging.cs
```diff
@@ -127,48 +102,31 @@ public async Task Toggle()
...
+                    {
+                        ServerConfig.IsOn = true;
+                        await Context.Channel.SendMessageAsync("Logger Enabled!");
+                    }
+                    else if (bEnabled)
+                    {
+                        ServerConfig.IsOn = false;
+                        await Context.Channel.SendMessageAsync("Logger Disabled!");
+                    }
+                    else // should never happen, but incase it does, reset the value
```
jkchen2, line 118: Does VS not give you a warning about this? If its something that should never happen (which it certainly appears to be the case), I'd just remove it altogether.

Floofbot/Modules/Logging.cs
```diff
@@ -177,226 +135,297 @@ public async Task Toggle()
...
-                dbConnection.Open();
+                _floofDb = floofDb;
             }
-            public async Task<ITextChannel> GetChannel(string tableName, Discord.IGuild guild)
+            public async Task<ITextChannel> GetChannel(string eventName, Discord.IGuild guild)
             {
-                // gets a channel based on the tablename (the type of logger)
-                string sqlGetChannel = $"SELECT {tableName} FROM Logger WHERE ServerID = {guild.Id} LIMIT 1"; // check state
-                using (SqliteCommand command = new SqliteCommand(sqlGetChannel, dbConnection))
+                // TODO: Find a better algorithm for this. Hardcoding event names is :(
```
jkchen2, line 145: Instead of having to add a `case` for each event type, you can instead try to get the channel ID by treating the `serverConfig` like a map. See [here](https://stackoverflow.com/questions/11395315) on getting property values from arbitrary key names.

Floofbot/Modules/Logging.cs
```diff
@@ -177,226 +135,297 @@ public async Task Toggle()
...
-                                return false;
-                        }
-                        return false;
-                    }
-                }
+                var ServerConfig = _floofDb.LogConfigs.Find(guild.Id);
+                if (ServerConfig == null) // no entry in DB for server - not configured
+                    return false;
+
+                bool bEnabled = ServerConfig.IsOn;
```
jkchen2, line 213: :foxsockangry:

Floofbot/Modules/Logging.cs
iShauny: this is legacy code from before this was created as a bool, I will delete this

Floofbot/Modules/Logging.cs
iShauny: whats the issue with this?

Floofbot/Modules/Logging.cs
iShauny: I agree on the validEvents part and will update the code to reflect this. I am confused by the first part, could you give an example of what you mean? I don't think I have ever used "maps"

jkchen2: Maps being hashmaps or something like Python dictionaries, or JS objects if you're familiar. Essentially, the goal is to have some arbitrary string key and have it map to a value on the `serverConfig` object.

Say we have an arbitrary `eventName` as an input:
```cs
string eventName = "MessageDeletedChannel";
var serverConfig = _floofDb.LogConfigs.Find(guild.Id);
System.Reflection.PropertyInfo propertyInfo = serverConfig.GetType().GetProperty(eventName);
ulong logChannel = (ulong)(propertyInfo.GetValue(serverConfig, null));
```

(Turns out the `null` in `GetValue` is [not for setting a default](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.propertyinfo.getvalue?view=netcore-3.1#System_Reflection_PropertyInfo_GetValue_System_Object_System_Object___). However, for the same reasoning as the second bit in the original comment, I think it should be fine.)

Floofbot/Modules/Logging.cs
jkchen2: This line can be changed to `return ServerConfig.IsOn;`, and the remaining lines can be removed. Again, seems like an artifact of previous design patterns.

Although having never worked with entity frameworks before, you may need to employ a cast to `bool` to keep the type checker happy.

iShauny: oh I see, Agreed. Will do this.

Floofbot/Modules/Logging.cs
iShauny: Oh I see what you are doing. Yeah, I think I was trying to just compensate for the possibility of a malformed input, but since its a function used by the developers only, I can see why this solution works.

PR #7, opened 2020-05-01
Carried out the migration

JasperGray:
- Renamed NicknameAlert in Services to NicknameAlertService to avoid name
collision.
- Ran the migration command.
  This also caused some previous table column names and/or types to change.
- Took off the partial keywords from each class in Models.
  This action does not affect the migration.

### DIFF
--- a/Floofbot/Migrations/20200501123654_AddNicknameAlertAndRenameCols.cs
+++ b/Floofbot/Migrations/20200501123654_AddNicknameAlertAndRenameCols.cs
@@ -1,0 +1,115 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Floofbot.Migrations
+{
+    public partial class AddNicknameAlertAndRenameCols : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.RenameColumn(
+                name: "UserID",
+                table: "Tags",
+                newName: "UserId");
+
+            migrationBuilder.RenameColumn(
+                name: "TagID",
+                table: "Tags",
+                newName: "TagId");
+
+            migrationBuilder.RenameColumn(
+                name: "ServerID",
+                table: "LogConfigs",
+                newName: "ServerId");
+
+            migrationBuilder.AlterColumn<ulong>(
+                name: "Id",
+                table: "Warnings",
+                nullable: false,
+                oldClrType: typeof(ulong),
+                oldType: "INTEGER")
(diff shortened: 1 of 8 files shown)


Floofbot/Migrations/20200501123654_AddNicknameAlertAndRenameCols.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    public partial class AddNicknameAlertAndRenameCols : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.RenameColumn(
+                name: "UserID",
+                table: "Tags",
+                newName: "UserId");
+
```
iShauny, line 13: Does the migration really need to modify the tag table?

Floofbot/Migrations/FloofDataContextModelSnapshot.cs
```diff
@@ -18,8 +18,12 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("Floofbot.Services.Repository.Models.LogConfig", b =>
                 {
+                    b.Property<ulong>("ServerId")
+                        .ValueGeneratedOnAdd()
+                        .HasColumnType("INTEGER");
+
                     b.Property<bool>("IsOn")
-                        .HasColumnType("INT");
+                        .HasColumnType("INTEGER");
```
iShauny, line 26: why is it modifying the logconfig model?

Floofbot/Migrations/20200501123654_AddNicknameAlertAndRenameCols.cs
JasperGray: The migration generates updates based on the names of the fields, so if the name of the field was UserID, but is now UserId, then it'll rename the Database.

Floofbot/Migrations/FloofDataContextModelSnapshot.cs
JasperGray: It's recognizing which fields are primary keys, which explains the HasKey calls.
I'm not sure why it's changing the type from Int to Integer, though.

Floofbot/Migrations/20200501123654_AddNicknameAlertAndRenameCols.cs
iShauny: have you experienced any issues with the other modules as a result of the changes?

JasperGray:
I'll temporarily close this PR because it seems the model fields need more changing than anticipated.

PR #8, opened 2020-05-01
Add a config file to the bot

JasperGray:
The config file contains the following:
- The prefix to be used for the bot, which is now a string
- The bot's custom status

### DIFF
--- a/Floofbot/Configs/BotConfig.cs
+++ b/Floofbot/Configs/BotConfig.cs
@@ -1,0 +1,28 @@
+using Discord;
+
+class BotConfig {
+    public static readonly string BotPrefix = ".";
+
+    public class BotActivity : IActivity {
+        public string Name {
+            get {
+                return "🦉 Simulator";
+            }
+        }
+        public ActivityType Type {
+            get {
+                return ActivityType.Playing;
+            }
+        }
+        public ActivityProperties Flags {
+            get {
+                return ActivityProperties.Play;
+            }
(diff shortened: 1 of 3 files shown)


jkchen2:
This is a pretty big shift from your implementation, but I think it'd be nice to have the configuration file be a [JSON](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to#how-to-read-json-into-net-objects-deserialize)/[YAML](https://github.com/aaubry/YamlDotNet)/INI file.

While this current implementation works, it's not as user friendly as editing a more universal config file format, say, YAML (I'm pretty impartial to that format for these types of things). More importantly, a change to the configuration would not require a project rebuild.

iShauny:
> This is a pretty big shift from your implementation, but I think it'd be nice to have the configuration file be a [JSON](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to#how-to-read-json-into-net-objects-deserialize)/[YAML](https://github.com/aaubry/YamlDotNet)/INI file.
> 
> While this current implementation works, it's not as user friendly as editing a more universal config file format, say, YAML (I'm pretty impartial to that format for these types of things). More importantly, a change to the configuration would not require a project rebuild.

Any way you could convert this to a json config file @JasperGray ? I agree with jsh here

JasperGray:
Yep I found a better way. Closing this one.

PR #9, opened 2020-05-01
Split event logger service into it's own file and pass the client object instead of istantiating the logger in the main program file

iShauny:
I removed the class from the module section and put it under the list of services, also changed the program.cs to reflect this


Floofbot/Services/EventLoggerService.cs
```diff
@@ -0,0 +1,496 @@
...
+                if (channel == null)
+                    return;
+
+                var embed = new EmbedBuilder()
+                 .WithTitle($"⚠️ Message Edited | {after.Author.Username}")
+                 .WithColor(Color.DarkGrey)
+                 .WithDescription($"{after.Author.Mention} ({after.Author.Id}) has edited their message in {channel.Mention}!")
+                 .AddField("Before", messageBefore.Content)
+                 .AddField("After", after.Content)
+                 .WithFooter(DateTime.Now.ToString("yyyy-MM-dd HH:mm"));
```
jkchen2, line 81: This is something I should have commented on earlier, but we can use the embed's built-in timestamp field to add an automatically formatted timestamp depending on region for the client:
https://discord.foxbot.me/docs/api/Discord.EmbedBuilder.html?q=EmbedBuilder#Discord_EmbedBuilder_WithCurrentTimestamp

Floofbot/Services/EventLoggerService.cs
```diff
@@ -0,0 +1,496 @@
...
+                Discord.ITextChannel logChannel = await GetChannel(channel.Guild, "MessageEditedChannel");
+                if (channel == null)
+                    return;
+
+                var embed = new EmbedBuilder()
+                 .WithTitle($"⚠️ Message Edited | {after.Author.Username}")
+                 .WithColor(Color.DarkGrey)
+                 .WithDescription($"{after.Author.Mention} ({after.Author.Id}) has edited their message in {channel.Mention}!")
+                 .AddField("Before", messageBefore.Content)
+                 .AddField("After", after.Content)
```
jkchen2, line 80: For these fields, I'd suggest making them [inline](https://discord.foxbot.me/docs/api/Discord.EmbedBuilder.html?q=EmbedBuilder#Discord_EmbedBuilder_AddField_System_String_System_Object_System_Boolean_) to minimize the size of embeds. I'd also argue that we should remove the profile picture display from log events that don't have much to do with the profile picture, but in this case I don't think it actually changes the display length of the embed.

Floofbot/Services/EventLoggerService.cs
iShauny: Can change this

jkchen2:
image

Floofbot/Services/EventLoggerService.cs
iShauny: done

Floofbot/Services/EventLoggerService.cs
```diff
@@ -0,0 +1,503 @@
...
+
+                if ((IsToggled(channel.Guild)) == false) // not toggled on
+                    return;
+
+                Discord.ITextChannel logChannel = await GetChannel(channel.Guild, "MessageEditedChannel");
+                if (logChannel == null)
+                    return;
+
+                var embed = new EmbedBuilder()
+                 .WithTitle($"⚠️ Message Edited | {after.Author.Username}")
```
jkchen2, line 76: This may be annoying to change on all embeds, but I'd suggest adding the discriminator to the end of the username. If a Discord account is deleted, I believe their ID is lost. It would be easier to search for logs if we could use the typical `<username>#<discriminator>` format instead of just the user ID.

Floofbot/Services/EventLoggerService.cs
```diff
@@ -0,0 +1,562 @@
...
+            var ServerConfig = _floofDb.LogConfigs.Find(guild.Id);
+            if (ServerConfig == null) // no entry in DB for server - not configured
+                return false;
+
+            return ServerConfig.IsOn;
+        }
+
+        public Task MessageUpdated(Cacheable<IMessage, ulong> before, SocketMessage after, ISocketMessageChannel chan)
+        {
+            var _ = Task.Run(async () =>
```
jkchen2, line 55: I'm still not the biggest fan of the usage of `Task.Run` here (and also the return of `Task.CompletedTask`), but will let it go for now. I don't have much input as I'm very much a C# novice, so I'll just put in an issue instead (as you suggested).

Floofbot/Services/EventLoggerService.cs
```diff
@@ -0,0 +1,562 @@
...
+                    Discord.ITextChannel channel = await GetChannel(user.Guild, "MemberUpdatesChannel");
+                    if (channel == null) // no log channel set
+                        return;
+
+                    var embed = new EmbedBuilder();
+
+                    if (before.Username != after.Username)
+                    {
+                        embed.WithTitle($"👥 Username Changed | {user.Username}#{user.Discriminator}")
+                            .WithColor(Color.Purple)
```
jkchen2, line 353: General note on style: definitely not a PR blocker, but I'd like to point out that we should probably stick to and/or agree on a certain code style. The only thing here is that the continuation indentation is inconsistent with some others in this file.

iShauny: I will fix this - wasnt intentional!

PR #10, opened 2020-05-01
Refactor embed-related methods

JasperGray:


Floofbot/Modules/Tag.cs
```diff
@@ -7,191 +7,188 @@
...
-                dbConnection = new SqliteConnection(new SqliteConnectionStringBuilder {
-                    DataSource = "botdata.db"
-                }.ToString());
-                dbConnection.Open();
-            }
+            dbConnection = new SqliteConnection(new SqliteConnectionStringBuilder {
+                DataSource = "botdata.db"
+            }.ToString());
+            dbConnection.Open();
+        }
```
iShauny, line 26: Use entity framework not sqlite

Floofbot/Modules/Tag.cs
```diff
@@ -7,191 +7,188 @@
...
+        [Command("add")]
+        [Priority(1)]
+        public async Task Add()
+        {
+            await SendEmbed(CreateDescriptionEmbed($"💾 Usage: `tag add [name] [content]`"));
+        }
 
-                List<string> tags = new List<string>();
-                List<string> pages = new List<string>();
+        //TODO: Fix me
```
iShauny, line 65: comment :thonk:

Floofbot/Modules/Tag.cs
```diff
@@ -7,191 +7,188 @@
...
+        [Command("list")]
+        public async Task ListTags([Remainder] string Content = null)
+        {
+            string guildId = $"{Context.Guild.Id}";
+            string sql = $"SELECT tagID FROM Tags WHERE tagID LIKE '%:$guildId%'";
```
iShauny, line 70: I'm guessing you are putting the guildID as part of the tagID

I am not sure I like this approach, I think it would be better to have a column for server ID

Floofbot/Modules/Tag.cs
```diff
@@ -7,191 +7,188 @@
...
-                            text += $"[{index}] - {tags[index].ToLower()}\n";
-                        }
+            tags = tags.OrderBy(x => x).ToList();
+            int index = 0;
+            for (int i = 1; i <= (tags.Count / 50) + 1; i++) {
+                string text = "```glsl\n";
+                int pagebreak = index;
+                for (; index < pagebreak + 50; index++) {
+                    if (index < tags.Count) {
+                        text += $"[{index}] - {tags[index].ToLower()}\n";
```
iShauny, line 89: this seems a bit messy, from what I gather, couldnt you just use foreach?

Floofbot/Modules/Tag.cs
```diff
@@ -7,191 +7,188 @@
...
-                if (Convert.ToInt32(command.ExecuteScalar()) > 0) {
-                    string delete = @"DELETE FROM Tags WHERE TagID = $TagId";
-                    command = new SqliteCommand(delete, dbConnection);
-                    command.Parameters.Add(new SqliteParameter("$TagId", tagId));
-                    command.ExecuteScalar();
+            if (Convert.ToInt32(command.ExecuteScalar()) > 0) {
+                string delete = @"DELETE FROM Tags WHERE TagID = $TagId";
+                command = new SqliteCommand(delete, dbConnection);
+                command.Parameters.Add(new SqliteParameter("$TagId", tagId));
+                command.ExecuteScalar();
```
iShauny: Do you really need to count the rows here?

Floofbot/Modules/Tag.cs
```diff
@@ -7,191 +7,188 @@
...
-                    command.Parameters.Add(new SqliteParameter("$TagID", TagID));
-                    var result = command.ExecuteScalar();
+        [Command("remove")]
+        [Priority(1)]
+        public async Task Remove()
+        {
+            await SendEmbed(CreateDescriptionEmbed($"💾 Usage: `tag remove [name]`"));
+        }
+
+        [Command]
```
iShauny, line 130: No command name?

Floofbot/Modules/Tag.cs
JasperGray: This part will be handled by a future PR.

Floofbot/Modules/Tag.cs
JasperGray: That's correct. This case handles commands like ".tag ..."

iShauny: Oh I see now!

Floofbot/Modules/Tag.cs
JasperGray: I'd prefer handling this in a future PR that deals with the models and/or the entity framework integration.

Floofbot/Modules/Tag.cs
JasperGray: I'd prefer handling this in a future PR that deals with the models and/or the entity framework integration.

Floofbot/Modules/Tag.cs
```diff
@@ -4,215 +4,191 @@
...
 
-                        EmbedBuilder builder = new EmbedBuilder() {
-                            Description = $"💾 Added Tag `{Tag}`",
-                            Color = Color.Magenta
-                        };
-                        await Context.Channel.SendMessageAsync("", false, builder.Build());
+        [Command("add")]
+        [Priority(0)]
+        [RequireUserPermission(GuildPermission.AttachFiles)]
+        public async Task Add(string Tag, [Remainder] string Content = null)
```
jkchen2, line 31: So I know this is kind of an artifact from the previous iteration of this code, but the capitalized variable names are throwing me off a bit.

Floofbot/Modules/Tag.cs
```diff
@@ -4,215 +4,191 @@
...
-
-            public Tag()
-            {
-                dbConnection = new SqliteConnection(new SqliteConnectionStringBuilder {
-                    DataSource = "botdata.db"
-                }.ToString());
-                dbConnection.Open();
-            }
+        private SqliteConnection dbConnection;
+        private static readonly Discord.Color embedColor = Color.Magenta;
```
jkchen2, line 18: Optional style: constants should be in yelling snake case. Also, C# knowledge limitation, but would there be a functional difference between using `readonly` vs `const` here? It looks like we're not really taking advantage of the slight flexibility `readonly` has here.

Floofbot/Modules/Tag.cs
JasperGray: I'll handle this in a future PR as part of #14 .

Floofbot/Modules/Tag.cs
JasperGray: It's a bit of futureproofing. For the readonly vs const I'm following the last part of https://stackoverflow.com/a/56024

Floofbot/Modules/Tag.cs
JasperGray: Actually, yes. I tried logging the output of the result of the delete query and it was null.

JasperGray:
Rebased

Floofbot/Modules/Tag.cs
JasperGray: fixed

Floofbot/Modules/Tag.cs
JasperGray: As the for constant in screaming snake case, fixed

Floofbot/Modules/Tag.cs
```diff
@@ -4,215 +4,191 @@
...
+                command.Parameters.Add(new SqliteParameter("$TagID", TagID));
+                var result = command.ExecuteScalar();
+
+                if (result != null) {
+                    string tag = result.ToString().Replace("@", "[at]");
+
+                    bool isImage = false;
+                    if (Uri.IsWellFormedUriString(tag, UriKind.RelativeOrAbsolute)) {
+                        string ext = tag.Split('.').Last().ToLower();
+                        isImage = (
```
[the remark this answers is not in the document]
JasperGray: Fixed, but left the variable in the same place for now since it's only used at that exact spot.

PR #11, opened 2020-05-01
Add random responses

JasperGray:


Floofbot/Handlers/CommandHandler.cs
```diff
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Reflection;
 using System.Threading.Tasks;
+using Discord;
```
iShauny, line 4: do we need to import the whole discord module?

JasperGray: Yep. It's to ensure the MessageSource enum can be found.

PR #15, opened 2020-05-02
Add Token field to app.config

JasperGray:
- Add app.config to gitignore and delete it from the repository
- Create a sample app.config file so users know what fields are needed
- The program now doesn't take in a parameter for the token

### DIFF
--- a/Floofbot/Program.cs
+++ b/Floofbot/Program.cs
@@ -1,13 +1,12 @@
 using System;
-using Floofbot.Services;
+using System.Configuration;
 using System.Threading.Tasks;
 using Discord;
 using Discord.WebSocket;
-using Serilog;
 using Floofbot.Configs;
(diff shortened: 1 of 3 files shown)


Floofbot/Program.cs
```diff
@@ -21,17 +20,7 @@ static async Task Main(string[] args)
...
-            if (!(args.Length == 1))
-            {
-                Console.WriteLine("Enter Bot Token");
-                token = Console.ReadLine();
-            }
-            else
-            {
-                token = args[0];
-            }
+            string token = ConfigurationManager.AppSettings["Token"];
```
iShauny, line 23: maybe add a check here for a blank token and tell the user how to modify it?

JasperGray: good idea -- done

PR #17, opened 2020-05-02
Created Nickname Alert Commands + DB Migration

iShauny:


Floofbot/Floofbot.csproj
```diff
@@ -5,6 +5,11 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="Migrations\20200501113559_AddNicknameAlert.cs" />
```
JasperGray, line 9: What's this part doing?

Floofbot/Migrations/FloofDataContextModelSnapshot.cs
```diff
@@ -18,8 +18,12 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
             modelBuilder.Entity("Floofbot.Services.Repository.Models.LogConfig", b =>
                 {
+                    b.Property<ulong>("ServerId")
```
JasperGray, line 21: These parts aren't reflected in the migration, but the bot still runs after?
If so, are we essentially manually modifying the context to reflect the DB, but not recording it in a migration?
If both are true then I think it is fine for now, as we should be migrating off of SQLite anyway later on.

Floofbot/Modules/NicknameAlert.cs
```diff
@@ -0,0 +1,78 @@
+﻿using Discord.Commands;
+using Floofbot.Services.Repository;
+using Floofbot.Services.Repository.Models;
+using Serilog;
+using System;
+using System.Threading.Tasks;
+
+namespace Floofbot.Modules
+{
+    public class NicknameAlert : ModuleBase<SocketCommandContext>
```
JasperGray, line 10: I'd honestly prefer a naming convention of XModule or XCommands for these module classes, but we can address this in a future PR.

Floofbot/Modules/NicknameAlert.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    ServerId = server,
+                    Channel = 0,
+                    IsOn = false
+                }) ;
+                _floofDB.SaveChanges();
+            }
+        }
+
+        [Command("setchannel")] // update into a group
```
JasperGray, line 35: Wait a minute, are these commands not in a commandgroup?
It should be something like .nicknamealert setchannel right?

Floofbot/Services/NicknameAlert.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Discord;
+using Floofbot.Services.Repository;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Floofbot.Services
+{
+    public class NicknameAlert
```
JasperGray, line 9: I'd really prefer if this were renamed to NicknameAlertService.
I'm also not sure what this file is for.

Floofbot/Floofbot.csproj
iShauny: It deletes the migration file after the migration has been added to the database, not compulsory

Floofbot/Services/NicknameAlert.cs
iShauny: Can rename this in a future PR, this is just a placeholder for the service right now

Floofbot/Modules/NicknameAlert.cs
iShauny: fixed, sorry

Floofbot/Migrations/FloofDataContextModelSnapshot.cs
iShauny: both are true, but to be honest we will be generating a new migration for when we move onto mysql

Floofbot/Modules/NicknameAlert.cs
```diff
@@ -0,0 +1,79 @@
...
+        public async Task Toggle()
+        {
+
+            // try toggling
+            try
+            {
+                // check the status of logger
+                var ServerConfig = _floofDB.NicknameAlertConfigs.Find(Context.Guild.Id);
+
+                bool bEnabled = ServerConfig.IsOn;
```
jkchen2, line 57: Optional:
```suggestion
                ServerConfig.IsOn = !ServerConfig.IsOn;
                await Context.Channel.SendMessageAsync("Nickname Alerts " + (ServerConfig.IsOn ? "Enabled!" : "Disabled!"));
```

Otherwise, typo on `Nicknqme`

jkchen2:
Was adding `floofData.db` intentional?

PR #23, opened 2020-05-02
Re-implement config file management to use YAML

JasperGray:


Floofbot/Configs/BotConfigFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+        public static void Initialize(string filename)
+        {
+            _filename = filename;
+        }
+
+        class BotConfigParser
+        {
+            public static BotConfig ParseFromFile(string filename)
```
jkchen2, line 32: No error handling on missing file?

Floofbot/Handlers/RandomResponseGenerator.cs
```diff
@@ -1,32 +1,34 @@
...
+            if ((response.RequireExact && loweredMessageContent == loweredRequiredInput)
+                || loweredMessageContent.Contains(loweredRequiredInput))
+            {
+                if (val < response.Probability)
+                {
+                    return response.Response;
+                }
+                else
+                {
+                    break;
```
jkchen2, line 29: Should this be `continue`?

JasperGray: Oh true. It's possible that a higher-probability response that also matches comes later.

Floofbot/Configs/BotConfigFactory.cs
JasperGray: Fixed

PR #24, opened 2020-05-03
Implemented Help Command and Pagination

iShauny:
The current available pagination modules are outdated, so a local copy of the project has been imported and updated to work with discord.net v2.2.0. A basic help command has been implemented that will be refactored and improved in future PRs.

### DIFF
--- a/Floofbot/Floofbot.csproj
+++ b/Floofbot/Floofbot.csproj
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <ProjectReference Include="..\Discord.Addons.Interactive\Discord.Addons.Interactive.csproj" />
+    <ProjectReference Include="../Discord.Addons.Interactive/Discord.Addons.Interactive.csproj" />
     <PackageReference Include="Discord.Net" Version="2.2.0" />
     <PackageReference Include="Microsoft.Data.Sqlite" Version="3.1.3" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="3.1.3" />
--- a/Floofbot/Floofbot.csproj
+++ b/Floofbot/Floofbot.csproj
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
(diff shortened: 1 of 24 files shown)


Floofbot/Floofbot.csproj
```diff
@@ -7,7 +7,7 @@
 
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
-    <PackageReference Include="Discord.Addons.Interactive" Version="1.0.1" />
+    <ProjectReference Include="..\Discord.Addons.Interactive\Discord.Addons.Interactive.csproj" />
```
JasperGray, line 10: Change to / for Linux pls. Pretty sure Windows supports forward slashes for these paths, too.

Floofbot/Modules/Utilities.cs
```diff
@@ -16,7 +16,8 @@ public async Task Ping()
             sw.Stop();
             await msg.DeleteAsync();
 
-            EmbedBuilder builder = new EmbedBuilder() {
+            EmbedBuilder builder = new EmbedBuilder()
```
JasperGray, line 19: This file seems to only have formatting changes. Maybe include it in a separate PR or a separate commit so that it's easier to undo independently.
I'd also prefer if the newline was kept at the end of the file

Floofbot/Modules/Help.cs
```diff
@@ -0,0 +1,77 @@
...
+using Discord.Commands;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Floofbot.Modules
+{
+    public class Help : InteractiveBase
```
JasperGray, line 12: The help module and command can also get summaries

Floofbot/Floofbot.csproj
iShauny: done

Floofbot/Modules/Utilities.cs
iShauny: reverted, was autoformat by visual studio

Floofbot/Modules/Help.cs
iShauny: This will be addressed in a future PR

PR #27, opened 2020-05-03
Add xkcd command

JasperGray:


JasperGray:
> I'd suggest using the JSON API: https://xkcd.com/json.html

Done. Now we don't need HtmlAgilityPack either c:

PR #29, opened 2020-05-03
Completed The NicknameAlertService

iShauny:
The NicknameAlertService was completed and will be integrated with the filter and event logger in a future PR. Some mild bugs have also been fixed in related commands.

### DIFF
--- a/Floofbot/Services/NicknameAlertService.cs
+++ b/Floofbot/Services/NicknameAlertService.cs
@@ -162,8 +162,6 @@ namespace Floofbot.Services
                     }
                     alertMessageIdsDic.Remove(msg.Id);
                 }
-                else
-                    return;
                 return;
             }
         }
--- a/Floofbot/Services/NicknameAlertService.cs
+++ b/Floofbot/Services/NicknameAlertService.cs
@@ -15,7 +15,7 @@ namespace Floofbot.Services
     {
         private FloofDataContext _floofDb;
         
-        private Dictionary<ulong, SocketGuildUser> messageDic = new Dictionary<ulong, SocketGuildUser>(); 
+        private Dictionary<ulong, SocketGuildUser> alertMessageIdsDic = new Dictionary<ulong, SocketGuildUser>(); 
         private ITextChannel _channel;
         private static readonly Emoji BAN_EMOJI = new Emoji("🔨");
         private static readonly Emoji WARN_EMOJI = new Emoji("⚠️");
@@ -56,7 +56,7 @@ namespace Floofbot.Services
             await message.AddReactionAsync(WARN_EMOJI);
             await message.AddReactionAsync(BAN_EMOJI);
 
-            messageDic.Add(message.Id, badUser);
+            alertMessageIdsDic.Add(message.Id, badUser);
 
         }
 
@@ -64,10 +64,10 @@ namespace Floofbot.Services
         {
             var msg = message.Value as IUserMessage;
 
-            if (messageDic.ContainsKey(msg.Id))
+            if (alertMessageIdsDic.ContainsKey(msg.Id))
             {
                 SocketGuildUser badUser;
-                messageDic.TryGetValue(msg.Id, out badUser);
+                alertMessageIdsDic.TryGetValue(msg.Id, out badUser);
                 var moderator = badUser.Guild.GetUser(reaction.UserId);
 
                 if (reaction.Emote.Name.Equals(BAN_EMOJI.Name))
@@ -91,7 +91,7 @@ namespace Floofbot.Services
                         await channel.SendMessageAsync("Unable to ban user. Do I have the permissions?");
                         Log.Error("Unable to ban user for bad name: " + ex);
                     }
-                    messageDic.Remove(msg.Id);
+                    alertMessageIdsDic.Remove(msg.Id);
                     return;
                 }
                 if (reaction.Emote.Name.Equals(WARN_EMOJI.Name))
@@ -124,7 +124,7 @@ namespace Floofbot.Services
                         await channel.SendMessageAsync("Unable to warn user. Do I have the permissions?");
                         Log.Error("Unable to warn user for bad name: " + ex);
                     }
-                    messageDic.Remove(msg.Id);
+                    alertMessageIdsDic.Remove(msg.Id);
                     return;
                 }
                 if (reaction.Emote.Name.Equals(KICK_EMOJI.Name))
@@ -148,7 +148,7 @@ namespace Floofbot.Services
                         await channel.SendMessageAsync("Unable to kick user. Do I have the permissions?");
                         Log.Error("Unable to kick user for bad name: " + ex);
                     }
-                    messageDic.Remove(msg.Id);
+                    alertMessageIdsDic.Remove(msg.Id);
                     return;
                 }
                 if (reaction.Emote.Name.Equals(REMOVE_EMOJI.Name))
@@ -163,7 +163,7 @@ namespace Floofbot.Services
                         await channel.SendMessageAsync("Unable to remove their nickname. Do I have the permissions?");
                         Log.Error("Unable to remove nickname for bad name: " + ex);
                     }
-                    messageDic.Remove(msg.Id);
+                    alertMessageIdsDic.Remove(msg.Id);
                     return;
                 }
                 else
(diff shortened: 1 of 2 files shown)


Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+            _channel = await GetChannel(guild, serverConfig.Channel);
+
+            var embed = new EmbedBuilder()
+                .WithDescription($"{removeEmoji.Name}: Remove Nickname\n" +
+                $"{warnEmoji.Name}: Warn\n" +
+                $"{kickEmoji.Name}: Kick\n" +
+                $"{banEmoji.Name}: Ban")
+                .Build();
+
+            var message = await _channel.SendMessageAsync($"{badUser.Mention} ({badUser.Username}#{badUser.Discriminator}) has been " +
```
JasperGray, line 51: Does this work if even their username is bad? Or does it only work once they set a nickname?

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+                $"detected with a bad name! What should I do?\n\nNickname: {badUser.Nickname}", false, embed);
+            await message.AddReactionAsync(removeEmoji);
+            await message.AddReactionAsync(kickEmoji);
+            await message.AddReactionAsync(warnEmoji);
+            await message.AddReactionAsync(banEmoji);
+
+            messageDic.Add(message.Id, badUser);
```
JasperGray, line 58: So messageDic is being used as temp storage?
The variable name can probably be changed to reflect that.
I'm thinking badNicknameMessageUserStore but it still sounds eh.

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+        }
+
+        public async Task ReactionCallback(Cacheable<IUserMessage, ulong> message, ISocketMessageChannel channel, SocketReaction reaction)
+        {
+            var msg = message.Value as IUserMessage;
+
+            if (messageDic.ContainsKey(msg.Id))
+            {
+                SocketGuildUser badUser;
+                messageDic.TryGetValue(msg.Id, out badUser);
```
JasperGray, line 69: Is this guaranteed to work? If so, just calling GetValue would be clearer

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+            await message.AddReactionAsync(removeEmoji);
+            await message.AddReactionAsync(kickEmoji);
+            await message.AddReactionAsync(warnEmoji);
+            await message.AddReactionAsync(banEmoji);
+
+            messageDic.Add(message.Id, badUser);
+
+        }
+
+        public async Task ReactionCallback(Cacheable<IUserMessage, ulong> message, ISocketMessageChannel channel, SocketReaction reaction)
```
JasperGray, line 62: Would be clearer if it was like an event name such as OnReactionPressed

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+                    try
+                    {
+                        //sends message to user
+                        EmbedBuilder builder = new EmbedBuilder();
+                        builder.Title = "⚖️ Ban Notification";
+                        builder.Description = $"You have been banned from {badUser.Guild.Name}";
+                        builder.AddField("Reason", "Banned by BOT for an inappropriate name");
+                        builder.Color = Color.DarkOrange;
+                        await badUser.SendMessageAsync("", false, builder.Build());
+
```
JasperGray, line 83: No need for blank lines here

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+                    }
+                    catch (Exception ex)
+                    {
+                        await channel.SendMessageAsync("Unable to ban user. Do I have the permissions?");
+                        Log.Error("Unable to ban user for bad name: " + ex);
+                    }
+                    messageDic.Remove(msg.Id);
+                    return;
+                }
+                if (reaction.Emote.Name.Equals(warnEmoji.Name))
```
JasperGray, line 96: else if instead of if can be used here, and then you don't have to write the return statements
Also, if the Name variables are both strings then just use ==

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+                    }
+                    catch (Exception ex)
+                    {
+                        await channel.SendMessageAsync("Unable to kick user. Do I have the permissions?");
+                        Log.Error("Unable to kick user for bad name: " + ex);
+                    }
+                    messageDic.Remove(msg.Id);
+                    return;
+                }
+                if (reaction.Emote.Name.Equals(removeEmoji.Name))
```
JasperGray, line 153: removeNicknameEmoji makes more sense here, since I thought it was remove as in removeUser at a glance.
Also, what if the removed nickname is even more offensive?

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+                    catch (Exception ex)
+                    {
+                        await channel.SendMessageAsync("Unable to remove their nickname. Do I have the permissions?");
+                        Log.Error("Unable to remove nickname for bad name: " + ex);
+                    }
+                    messageDic.Remove(msg.Id);
+                    return;
+                }
+                else
+                    return;
```
JasperGray, line 169: This else case is unnecessary -- the method will return by default after reaching the end.

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,174 @@
...
+        private Emoji banEmoji = new Emoji("🔨");
+        private Emoji warnEmoji = new Emoji("⚠️");
+        private Emoji kickEmoji = new Emoji("👢");
+        private Emoji removeEmoji = new Emoji("📝");
+
+        public NicknameAlertService(FloofDataContext floofDb)
+        {
+            _floofDb = floofDb;
+        }
+        public async Task<ITextChannel> GetChannel(Discord.IGuild guild, ulong channelId)
```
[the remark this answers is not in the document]
iShauny: getchannel can be private, handlebadnickname is called outwith the instance and has to remain public

Floofbot/Services/NicknameAlertService.cs
iShauny: I added in a ternary operator where I think an issue would occur, the rest of the code should be ok

Floofbot/Services/NicknameAlertService.cs
iShauny: Renamed to alertMessageIdsList

Floofbot/Services/NicknameAlertService.cs
iShauny: the event in the api is called ReactionAdded so renamed to OnReactionAdded

Floofbot/Services/NicknameAlertService.cs
iShauny: I tried using the == operator but ran into issues, so will leave this for now

Floofbot/Services/NicknameAlertService.cs
iShauny: There will be no nickname so I dont understand what you mean by this.

Floofbot/Services/NicknameAlertService.cs
iShauny: I can remove the else, but I would like to leave a return statement at the end, just incase of future expansion

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -0,0 +1,170 @@
...
+                        await channel.SendMessageAsync($"Got it! I changed {badUser.Username}#{badUser.Discriminator}'s nickname!");
+                    }
+                    catch (Exception ex)
+                    {
+                        await channel.SendMessageAsync("Unable to remove their nickname. Do I have the permissions?");
+                        Log.Error("Unable to remove nickname for bad name: " + ex);
+                    }
+                    alertMessageIdsDic.Remove(msg.Id);
+                }
+                return;
```
JasperGray, line 165: Is this return statement actually needed still?

PR #30, opened 2020-05-03
added minesweeper command

bealsbe:
//here be dragons

### DIFF
--- a/Floofbot/Modules/Fun.cs
+++ b/Floofbot/Modules/Fun.cs
@@ -168,5 +168,153 @@ namespace Floofbot.Modules
             }
             await Context.Channel.SendMessageAsync(fact);
         }
+
+        [Command("minesweeper")]
+        [Summary("Responds with a random cat fact")]
+        public async Task Minesweeper(int gridx, int gridy, int bombs)
+        {
+            Gameboard game = new Gameboard(gridx, gridy, bombs);
+            EmbedBuilder builder = new EmbedBuilder();
+            builder.Title = ":bomb: Minesweeper";
+            builder.Color = Discord.Color.Magenta;
+            builder.Description = game.getBoard();
+
+            await Context.Channel.SendMessageAsync("", false, builder.Build());
+        }
+
+
+        public class Gameboard
+        {
+            private string[,] grid;
+            private int x;
+            private int y;
+            private int bombCount;
+
+            public Gameboard(int _x, int _y, int _bombCount)
+            {
+                x = _x;
+                y = _y;
+                bombCount = _bombCount;
+                grid = new string[x, y];
+                plantBombs();
+
+
+                for (int i = 0; i < x; i++)
+                {
+                    for (int j = 0; j < y; j++)
+                    {
+                        if (grid[i, j] != "||:bomb:||")
+                        {
+                            grid[i, j] = getBombCount(i, j).ToString();
+                        }
+                    }
+                }
+            }
+
+            public int getBombCount(int _x, int _y)
+            {
+                int count = 0;
+
+                for (int i = -1; i <= 1; i++)
+                {
+                    for (int j = -1; j <= 1; j++)
+                    {
+                        if (!(i == 0 && j == 0))
+                        {
+                            try
+                            {
+                                if (grid[_x + i, _y + j] == "||:bomb:||")
+                                {
+                                    count++;
+                                }
+                            }
+                            catch { } //stupid but works I  guess
+                        }
+                    }
+                }
+
+                return count;
+            }
+
+            public void plantBombs()
+            {
+                Random rnd = new Random();
+
+                for (int i = 0; i < bombCount; i++)
+                {
+                    bool isGood = false;
+                    while (!isGood)
+                    {
+                        int randX = rnd.Next(0, x);
+                        int randY = rnd.Next(0, y);
+
+                        if (grid[randX, randY] != "||:bomb:||")
+                        {
+                            isGood = true;
+                            grid[randX, randY] = "||:bomb:||";
+                        }
+                    }
+                }
+            }
+            public string getBoard()
+            {
+                string board = "";
+                for (int i = 0; i < x; i++)
+                {
+                    for (int j = 0; j < y; j++)
+                    {
+                        if (grid[i, j] == "1")
+                        {
+                            board += "||:one:||";
+                        }
+                        else if (grid[i, j] == "2")
+                        {
+                            board += "||:two:||";
+                        }
+                        else if (grid[i, j] == "3")
+                        {
+                            board += "||:three:||";
+                        }
+                        else if (grid[i, j] == "4")
+                        {
+                            board += "||:four:||";
+                        }
+                        else if (grid[i, j] == "5")
+                        {
+                            board += "||:five:||";
+                        }
+                        else if (grid[i, j] == "6")
+                        {
+                            board += "||:six:||";
+                        }
+                        else if (grid[i, j] == "7")
+                        {
+                            board += "||:seven:||";
+                        }
+                        else if (grid[i, j] == "8")
+                        {
+                            board += "||:eight:||";
+                        }
+                        else if (grid[i, j] == "9")
+                        {
+                            board += "||:nine:||";
+                        }
+                        else if (grid[i, j] == "0")
+                        {
+                            board += "||:zero:||";
+                        }
+                        else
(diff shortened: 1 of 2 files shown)


Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,153 @@ public async Task RequestCatFact()
             }
             await Context.Channel.SendMessageAsync(fact);
         }
+
+        [Command("minesweeper")]
+        [Summary("Responds with a random cat fact")]
```
bealsbe, line 173: oops

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
...
+        }
+
+        public class Gameboard
+        {
+            private string[,] grid;
+            private int x;
+            private int y;
+            private int bombCount;
+
+            public Gameboard(int _x, int _y, int _bombCount)
```
iShauny, line 212: Usually the convention is the opposite
i.e.
private int _x;
private int _y;
_x = x; etc 
Not PR blocking problem though

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
...
+                        {
+                            try
+                            {
+                                if (grid[_x + i, _y + j] == "||:bomb:||")
+                                {
+                                    count++;
+                                }
+                            }
+                            catch { } //stupid but works I  guess
+                        }
```
iShauny, line 251: does this need to be in a try catch block? you can also use if, else continue;

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
...
+                return count;
+            }
+
+            public void plantBombs()
+            {
+                Random rnd = new Random();
+
+                for (int i = 0; i < bombCount; i++)
+                {
+                    bool isGood = false;
```
iShauny, line 264: I would maybe rename this something else like bombSpaceCandidate or something

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
...
+                        else if (grid[i, j] == "8")
+                        {
+                            board += "||:eight:||";
+                        }
+                        else if (grid[i, j] == "9")
+                        {
+                            board += "||:nine:||";
+                        }
+                        else if (grid[i, j] == "0")
+                        {
```
iShauny, line 322: could use case/switch or a dictionary

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
...
+            Gameboard game = new Gameboard(gridx, gridy, bombs);
+            EmbedBuilder builder = new EmbedBuilder();
+            builder.Title = ":bomb: Minesweeper";
+            builder.Color = EMBED_COLOR;
+            builder.Description = game.getBoard();
+
+            await Context.Channel.SendMessageAsync("", false, builder.Build());
+        }
+
+        public class Gameboard
```
iShauny, line 205: I don't think it needs to be a public class

JasperGray: lol I'd actually prefer if this were its own class in its own file.
Also, I'd prefer the capitalization GameBoard*
I don't think we currently have a place to put this class, but maybe Floofbot/Activities/GameBoard.cs works.
Then, each individual activity could become a class, since Fun.cs will soon become a god object.

Floofbot/Modules/Fun.cs
JasperGray: Yes, because this is doing a naive out of bounds check on the grid. I think it's worth commenting that rather than just calling it something stupid, though.

Floofbot/Modules/Fun.cs
JasperGray: isValidBombSpaceCandidate*, then
I like bool variables to take on names with the form hasX or isX in most cases

Floofbot/Modules/Fun.cs
JasperGray: definitely a dictionary, pls. Then we can do
`board += dictionaryName[grid[i, j]]`

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
...
+        {
+            private string[,] grid;
+            private int x;
+            private int y;
```
JasperGray, line 209: x and y usually refer to a specific point
pls rename these vars width and height

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +168,172 @@ public async Task RequestCatFact()
             }
             await Context.Channel.SendMessageAsync(fact);
         }
+
+        [Command("minesweeper")]
+        [Summary("Minesweeper minigame")]
+        public async Task Minesweeper([Summary("grid x")]int gridx, [Summary("grid y")]int gridy, [Summary("bomb count")]int bombs)
```
JasperGray, line 174: What happens if it's not an int? Is it just logged as an error and the bot acts like nothing happened?
I've been assuming the args had to be strings for some reason.

JasperGray, line 174: Yeah these parameters will have to be renamed width and height imo, rather than just gridx gridy

Floofbot/Modules/Fun.cs
iShauny: I think I would prefer /Modules/Helpers/GameBoard.cs if we do this approach

Floofbot/Modules/Fun.cs
iShauny: In that case I think I would prefer making the boundaries a variable that can be referenced than using errors

Floofbot/Modules/Fun.cs
bealsbe: I needs to be public otherwise it will error out with the class being inaccessible

Floofbot/Modules/Fun.cs
bealsbe: The bot logs an error and it will not respond to the command

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +169,40 @@ public async Task RequestCatFact()
...
+                return;
+            }
+            
+            //limits the size of the board
+            if(width > 10 || height > 10)
+            {
+                await Context.Channel.SendMessageAsync("Max Grid Size: 10 x 10");
+                return;
+            }
+            else if(bombs >= height * width)
```
JasperGray, line 191: Not PR blocking, but if you're going to return from an if statement, then you can leave this as another if statement, rather than an else-if.

Floofbot/Modules/Fun.cs
```diff
@@ -168,5 +169,40 @@ public async Task RequestCatFact()
...
+            {
+                await Context.Channel.SendMessageAsync("Too many bombs!");
+                return;
+            }
+
+            GameBoard game = new GameBoard(height, width, bombs);
```
JasperGray, line 197: Sanity check: width and height get switched, here? 
Because the command takes in the width first, followed by the height.

Floofbot/Modules/Helpers/GameBoard.cs
```diff
@@ -0,0 +1,109 @@
...
+            for (int i = -1; i <= 1; i++)
+            {
+                for (int j = -1; j <= 1; j++)
+                {
+                    if (!(i == 0 && j == 0))
+                    {
+                        //checks to see if the array goes out of bounds
+                        if (!(x + i < 0 || y + j < 0 || x + i >= grid.GetLength(0) || y + j >= grid.GetLength(1)))
+                        {
+                            if (grid[x + i, y + j] == "||:bomb:||")
```
JasperGray, line 47: Not PR blocking: you can reform these two if statements as follows:
```cs
if (x + i >= 0 
    && y + j >= 0 
    && x + i < grid.GetLength(0) 
    && y + j < grid.GetLength(1) 
    && grid[x + i, y + j] == "||:bomb:||")
{
}
```
because the if statements "short-circuit", meaning if an earlier condition is false then it doesn't check the later conditions. Also, changing the or's to and's along with reversing the conditions is by De Morgan's Laws.

PR #31, opened 2020-05-03
Refactor Tag model

JasperGray:
- Add column TagName of type string, so that TagId can remain an int
- Rename Content to TagContent

### DIFF
--- a/Floofbot/Migrations/20200503224258_AlterTagTable.cs
+++ b/Floofbot/Migrations/20200503224258_AlterTagTable.cs
@@ -1,0 +1,22 @@
+﻿using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace Floofbot.Migrations
+{
+    public partial class AlterTagTable : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "TagName",
+                table: "Tags",
+                nullable: true);
+        }
+
+        protected override void Down(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.DropColumn(
+                name: "TagName",
+                table: "Tags");
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


Floofbot/Migrations/20200503224258_AlterTagTable.cs
```diff
@@ -0,0 +1,22 @@
...
+namespace Floofbot.Migrations
+{
+    public partial class AlterTagTable : Migration
+    {
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.AddColumn<string>(
+                name: "TagName",
+                table: "Tags",
+                nullable: true);
```
iShauny, line 12: nullable = true?

Floofbot/Services/Repository/Models/Tag.cs
```diff
@@ -6,7 +6,8 @@ public partial class Tag
     {
         [Key]
         public ulong TagId { get; set; }
+        public string TagName { get; set; }
         public ulong UserId { get; set; }
```
iShauny, line 10: what's the purpose of user id? I think server id should be logged somewhere too, so that tags are not global

Floofbot/Migrations/20200503231617_AlterTagTable.cs
```diff
@@ -6,14 +6,25 @@ public partial class AlterTagTable : Migration
     {
         protected override void Up(MigrationBuilder migrationBuilder)
         {
+            migrationBuilder.AddColumn<ulong>(
+                name: "ServerId",
+                table: "Tags",
+                nullable: false,
+                defaultValue: 0ul);
```
iShauny, line 13: default value 0ul?

JasperGray: It means 0, but specifically unsigned long.

Floofbot/Services/Repository/Models/Tag.cs
JasperGray: Added a ServerId field.
I kept the UserId field so it stays clear who added what.

PR #32, opened 2020-05-03
Added say command

bealsbe:


Floofbot/Modules/Utilities.cs
```diff
@@ -67,14 +71,26 @@ public async Task Avatar([Remainder] IGuildUser user = null)
...
                 ImageUrl = avatarUrl,
-                Color = Color.Magenta
+                Color = EMBED_COLOR
 
             };
             await Context.Channel.SendMessageAsync("", false, builder.Build());
+        }
 
+        [Command("say")]
+        [Summary("Reapeats a message")]
```
iShauny, line 85: typo

Floofbot/Modules/Utilities.cs
```diff
@@ -67,14 +71,26 @@ public async Task Avatar([Remainder] IGuildUser user = null)
...
+        public async Task RepeatMessage([Remainder] string message =null)
+        {
+            EmbedBuilder builder = new EmbedBuilder()
+            {
+                Description = message,
```
iShauny, line 90: I would handle null messages. Also wary of @ tags in embeds.

bealsbe: The message is embed.  @everyone and other pings do not go through

Floofbot/Modules/Utilities.cs
```diff
@@ -85,12 +85,19 @@ public async Task Avatar([Remainder] IGuildUser user = null)
...
+                EmbedBuilder builder = new EmbedBuilder()
+                {
+                    Description = message,
+                    Color = EMBED_COLOR
+                };
+                await Context.Channel.SendMessageAsync("", false, builder.Build());
+            }
+            else
+            {
+                await Context.Channel.SendMessageAsync("Usage: `.say [message]);
```
iShauny, line 99: missing: `"

unexpected EOF

bealsbe: whoops

PR #33, opened 2020-05-04
Added A Check To Not Post Bot Messages In The Event Logger

iShauny:
The event logger service is updated to not post bot messages

### DIFF
--- a/Floofbot/Services/EventLoggerService.cs
+++ b/Floofbot/Services/EventLoggerService.cs
@@ -1,4 +1,5 @@
 ﻿using Discord;
+using Discord.Commands;
 using Discord.WebSocket;
 using Floofbot.Services.Repository;
 using Serilog;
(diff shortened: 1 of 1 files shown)


Floofbot/Services/EventLoggerService.cs
```diff
@@ -337,6 +361,9 @@ public Task GuildMemberUpdated(SocketGuildUser before, SocketGuildUser after)
             {
                 try
                 {
+                    if (after.IsBot)
```
JasperGray, line 364: Shouldn't this be placed after the null check?
Also, do we have to check for whether before is a bot, too?

iShauny: Moved it before it. No need because a user cannot change from a normal account to a bot account and vice versa. Before is just the user before they are updated

PR #34, opened 2020-05-04
Added Command For Specific Module Help

iShauny:
A command has been added that returns module-specific help, detailing parameters

Summaries have also been added

### DIFF
--- a/Floofbot/Modules/Help.cs
+++ b/Floofbot/Modules/Help.cs
@@ -9,6 +9,7 @@ using System.Threading.Tasks;
 
 namespace Floofbot.Modules
 {
+    [Summary("Commands for Listing Available Commands")]
     public class Help : InteractiveBase
     {
         private readonly IServiceProvider _services;
@@ -19,6 +20,7 @@ namespace Floofbot.Modules
             _commands = commands;
         }
 
+        [Summary("Show all available commands")]
         [Command("help")]
         public async Task HelpCommand()
         {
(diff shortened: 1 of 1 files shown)


Floofbot/Modules/Help.cs
```diff
@@ -73,5 +75,74 @@ public async Task HelpCommand()
...
+                List<string> moduleNames = new List<string>();
+                List<CommandInfo> commands = _commands.Commands.ToList();
+                List<EmbedFieldBuilder> fields = new List<EmbedFieldBuilder>();
+                List<PaginatedMessage.Page> pages = new List<PaginatedMessage.Page>();
+
+                // put available module names into a list instead of their object
+                foreach (ModuleInfo module in _commands.Modules.ToList())
+                    moduleNames.Add(module.Name.ToLower());
+
+                if (!moduleNames.Contains(requestedModule.ToLower())) // no such cmd
```
JasperGray, line 94: We can probably save a few lines here using .Any() from LINQ if you wanna be fancy. 
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.any?view=netcore-3.1#System_Linq_Enumerable_Any__1_System_Collections_Generic_IEnumerable___0__System_Func___0_System_Boolean__
What we have here is fine, though

Floofbot/Modules/Help.cs
```diff
@@ -73,5 +75,74 @@ public async Task HelpCommand()
                 Trash = true
             });
         }
+
+        [Summary("Show help for a specific module")]
+        [Command("help")]
+        public async Task HelpCommand([Summary("module")]string requestedModule)
+        {
+            try
```
JasperGray, line 83: Not sure why we need this try catch here.

Floofbot/Modules/Help.cs
```diff
@@ -73,5 +75,74 @@ public async Task HelpCommand()
...
+                foreach (ModuleInfo module in _commands.Modules.ToList())
+                    moduleNames.Add(module.Name.ToLower());
+
+                if (!moduleNames.Contains(requestedModule.ToLower())) // no such cmd
+                {
+                    await Context.Channel.SendMessageAsync("Unable to find that command");
+                    return;
+                }
+
+                foreach (CommandInfo cmd in commands)
```
JasperGray, line 100: This part can be refactored in a future PR -- we have some duplicated code going on.

Floofbot/Modules/Help.cs
```diff
@@ -73,5 +75,74 @@ public async Task HelpCommand()
                 Trash = true
             });
         }
+
+        [Summary("Show help for a specific module")]
+        [Command("help")]
+        public async Task HelpCommand([Summary("module")]string requestedModule)
```
JasperGray, line 81: Will we ever have a module longer than one word?
Just asking since there's currently no [Remainder] marker right now on the requestedModule param.

Floofbot/Modules/Help.cs
iShauny: I believe this can be refactored in a future PR if required

Floofbot/Modules/Help.cs
iShauny: left over from testing, my bad

Floofbot/Modules/Help.cs
iShauny: iirc module names cannot be multiple words - so no

Also this leaves the opportunity for even further commands detailing parameter details or even more sub commands  - would be done in a future pr if appropriate

PR #35, opened 2020-05-04
Remove manual SQLite in Tag module

JasperGray:


Floofbot/Modules/TagCommands.cs
```diff
@@ -0,0 +1,206 @@
...
+
+        private FloofDataContext _floofDb;
+
+        public TagCommands(FloofDataContext floofDb)
+        {
+            _floofDb = floofDb;
+        }
+
+        [Command("add")]
+        [Priority(0)]
```
iShauny, line 35: is the priority tag needed if we are overloading the function anyway?

Floofbot/Modules/TagCommands.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            // await PagedReplyAsync(pages);
+        }
+
+        [Command("remove")]
+        [Summary("Removes a tag")]
+        [RequireUserPermission(GuildPermission.ManageMessages)]
+        public async Task Remove([Summary("Tag name")] string tag)
+        {
+            string tagName = $"{tag.ToLower()}:{Context.Guild.Id}";
```
iShauny, line 113: do we still need to use guild id in the tag name if we are adding guild id to the db?

JasperGray: It honestly depends on how we'll migrate the data over from SQLite to MySQL.
If we can manually adjust the data before moving it over, that's nice.
This current approach is assuming that tagName is being used for everything right now, though.

iShauny: can be updated in a future pr i guess

Floofbot/Modules/TagCommands.cs
JasperGray: Just tested it. We need [Priority(1)] on the overload but we don't need [Priority(0)] here. I'll edit it out.

PR #38, opened 2020-05-04
Created Server Info Command

iShauny:


Floofbot/Modules/Utilities.cs
```diff
@@ -1,7 +1,12 @@
-﻿using System.Diagnostics;
+﻿using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics;
```
JasperGray, line 4: Do we really need all these includes?

Floofbot/Modules/Utilities.cs
```diff
@@ -99,5 +104,41 @@ public async Task RepeatMessage([Remainder] string message =null)
...
+
+        [Command("serverinfo")]
+        [Summary("Returns information about the current server")]
+        public async Task ServerInfo()
+        {
+            SocketGuild guild = Context.Guild;
+            int numberTextChannels = guild.TextChannels.Count;
+            int numberVoiceChannels = guild.VoiceChannels.Count;
+            string createdAt = $"Created {guild.CreatedAt.DateTime.ToShortDateString()}. " +
+                               $"That's over {Context.Message.CreatedAt.Subtract(guild.CreatedAt).Days} days ago!";
```
JasperGray, line 116: Be careful here. We'll probably need another case for if it ends up being 1 _day_ ago

Floofbot/Modules/Utilities.cs
```diff
@@ -99,5 +104,41 @@ public async Task RepeatMessage([Remainder] string message =null)
...
+            int totalMembers = guild.MemberCount;
+            int onlineUsers = guild.Users.Where(mem => mem.Status == UserStatus.Online).Count();
```
JasperGray, line 118: I think you can just use .Count(lambda) https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.count?view=netcore-3.1#System_Linq_Enumerable_Count__1_System_Collections_Generic_IEnumerable___0__System_Func___0_System_Boolean__

Floofbot/Modules/Utilities.cs
```diff
@@ -99,5 +104,41 @@ public async Task RepeatMessage([Remainder] string message =null)
...
+            int numberRoles = guild.Roles.Count;
+            int numberEmojis = guild.Emotes.Count;
+            uint colour = (uint)new Random().Next(0x1000000); // random hex
```
JasperGray, line 121: lol but why though

Floofbot/Modules/Utilities.cs
```diff
@@ -99,5 +104,41 @@ public async Task RepeatMessage([Remainder] string message =null)
...
+
+            EmbedBuilder embed = new EmbedBuilder();
+
+            embed.WithDescription(createdAt)
+                 .WithColor(new Discord.Color(colour))
+                 .AddField("Users", $"{onlineUsers}/{totalMembers}", true)
```
JasperGray, line 127: "Users (online/total)" would be clearer

Floofbot/Modules/Utilities.cs
iShauny: looks good

Floofbot/Modules/Utilities.cs
iShauny: I dont see much of a difference in
`guild.Users.Count(mem =>  mem.Status == userStatus.Online);`
and what I have put here

Floofbot/Modules/Utilities.cs
```diff
@@ -99,5 +102,44 @@ public async Task RepeatMessage([Remainder] string message =null)
...
+        [Summary("Returns information about the current server")]
+        public async Task ServerInfo()
+        {
+            SocketGuild guild = Context.Guild;
+            int numberTextChannels = guild.TextChannels.Count;
+            int numberVoiceChannels = guild.VoiceChannels.Count;
+            int daysOld = Context.Message.CreatedAt.Subtract(guild.CreatedAt).Days;
+            string createdAt = $"Created {guild.CreatedAt.DateTime.ToShortDateString()}. " +
+                               $"That's " +
+                               ((daysOld > 1) ? $"over {daysOld} days" : $"{daysOld} day") +
```
JasperGray, line 116: What about 0 _days_ now? :^) 
You might want to create a new string variable separately first to avoid createdAt being a long string full of ternaries.

PR #41, opened 2020-05-04
Added Filter Migration

iShauny:


Floofbot/Services/Repository/Models/FilteredWord.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Floofbot.Services.Repository.Models
+{
+    public partial class FilteredWord
+    {
+        [Key]
+        public ulong Id { get; set; }
+        public ulong ServerId { get; set; }
+        public string Word { get; set; }
```
JasperGray, line 10: ok so the wordlist is by server rather than by channel?

Floofbot/Services/Repository/Models/FilterConfig.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.ComponentModel.DataAnnotations;
+
+namespace Floofbot.Services.Repository.Models
+{
+    public partial class FilterConfig
+    {
+        [Key]
+        public ulong ServerId { get; set; }
+        public bool IsOn { get; set; }
```
JasperGray, line 9: I'm not sure what this table does when there's already FilterChannelWhiteList above.

Floofbot/Services/Repository/Models/FilteredWord.cs
iShauny: Yes, filtered words is per server

Floofbot/Services/Repository/Models/FilterConfig.cs
iShauny: This one enables/disables the filter for the whole server, the channel whitelists are channels that are ignored for the filter

PR #43, opened 2020-05-05
Added serverlist command (owner only)

iShauny:


Floofbot/Modules/Utilities.cs
```diff
@@ -139,5 +141,60 @@ public async Task ServerInfo()
...
+            List<PaginatedMessage.Page> pages = new List<PaginatedMessage.Page>();
+
+            foreach (SocketGuild g in guilds)
+            {
+                List<EmbedFieldBuilder> fields = new List<EmbedFieldBuilder>();
+
+                fields.Add(new EmbedFieldBuilder()
+                {
+                    Name = $"Owner",
+                    Value = $"{g.Owner.Username}#{g.Owner.Discriminator} | ``{g.Owner.Id}``",
```
JasperGray, line 159: Sanity check: is this the server owner or the bot owner?
Maybe change the name field to "Server Owner"

Floofbot/Modules/Utilities.cs
```diff
@@ -139,5 +141,60 @@ public async Task ServerInfo()
...
+                fields.Add(new EmbedFieldBuilder()
+                {
+                    Name = $"Members",
+                    Value = g.MemberCount,
+                    IsInline = false
+                });
+
+                pages.Add(new PaginatedMessage.Page
+                {
+                    Author = new EmbedAuthorBuilder { Name = g.Name },
```
JasperGray, line 177: Actually, what does Author here do?
Is it like the "Title" field?

Floofbot/Modules/Utilities.cs
```diff
@@ -139,5 +141,60 @@ public async Task ServerInfo()
...
+                Options = PaginatedAppearanceOptions.Default,
+                TimeStamp = DateTimeOffset.UtcNow
+            };
+            await PagedReplyAsync(pager, new ReactionList
+            {
+                Forward = true,
+                Backward = true,
+                Jump = true,
+                Trash = true
+            }, true);
```
JasperGray, line 197: What's this true parameter for?

Floofbot/Modules/Utilities.cs
```diff
@@ -1,14 +1,16 @@
...
+using Discord.Addons.Interactive;
 using Discord.Commands;
 using Discord.WebSocket;
 
 namespace Floofbot
 {
-    public class Utilities : ModuleBase<SocketCommandContext>
+    public class Utilities : InteractiveBase
     {
         private static readonly Discord.Color EMBED_COLOR = Color.Magenta;
```
JasperGray, line 15: We'll need to figure out these colours and where to place them in a future PR considering that the serverlist command has a dark green (admin) embed colour right now.

Floofbot/Modules/Utilities.cs
iShauny: Only the person who ran the command will have their reactions counted

Floofbot/Modules/Utilities.cs
iShauny: Kinda like the title per page

Floofbot/Modules/Utilities.cs
iShauny: Server owner

Floofbot/Modules/Utilities.cs
iShauny: Lol didn't even notice this I can swap it to the random colour?

PR #45, opened 2020-05-05
fixed issue with where warnlog would try to list warnings for users w…

bealsbe:


Floofbot/Modules/Administration.cs
```diff
@@ -156,11 +156,18 @@ public async Task warnlog([Summary("user")] string user)
...
+                await Context.Channel.SendMessageAsync($"{badUser.Username}#{badUser.Discriminator} is a good noodle. They have no warnings!");
+                return;
+            }
+
             EmbedBuilder builder = new EmbedBuilder();
             int warningCount = 0;
             builder.WithTitle($"Warnings for {badUser.Username}#{badUser.Discriminator}");
             foreach (Warning warning in warnings) {
                 builder.AddField($"**{warningCount + 1}**. {warning.DateAdded.ToString("yyyy-MM-dd")}", $"```{warning.Reason}```");
+                builder.Color = ADMIN_COLOR;
```
JasperGray, line 170: This can be outside of the loop

Floofbot/Modules/Administration.cs
```diff
@@ -156,7 +156,14 @@ public async Task warnlog([Summary("user")] string user)
...
 
+            if (warnings.Count() == 0)
+            {
+                await Context.Channel.SendMessageAsync($"{badUser.Username}#{badUser.Discriminator} is a good noodle. They have no warnings!");
+                return;
+            }
+
             EmbedBuilder builder = new EmbedBuilder();
+            builder.Color = ADMIN_COLOR;
             int warningCount = 0;
```
JasperGray, line 167: int warningCount should be declared before the builder is, since it's kinda confusing that there's a random int being declared between builder.Color and builder.WithTitle.
I'll let it slide though

PR #47, opened 2020-05-05
Add WordFilterService

JasperGray:


Floofbot/Services/EventLoggerService.cs
```diff
@@ -595,5 +602,37 @@ public Task UserUnmuted(IUser user, IUser unmuter, IGuild guild)
...
+
+        class WordFilterService
+        {
+            List<FilteredWord> _wordfilterList;
+
+            bool hasFilteredWord(FloofDataContext floofDb, string messageContent, ulong serverId, ulong channelId)
+            {
+                // whitelist means we don't have the filter on for this channel
+                if (floofDb.FilterChannelWhitelists.AsQueryable()
+                    .Any(x => x.ChannelId == channelId && x.ServerId == serverId))
```
iShauny, line 614: also need to check FilterConfigs either if server has not been set uo (no entry) or disabled on whole server (IsOn = false)

Floofbot/Services/EventLoggerService.cs
```diff
@@ -595,5 +602,37 @@ public Task UserUnmuted(IUser user, IUser unmuter, IGuild guild)
...
+                {
+                    return false;
+                }
+
+                if (_wordfilterList == null)
+                {
+                    _wordfilterList = floofDb.FilteredWords.AsQueryable()
```
iShauny, line 621: only issue I see is if we add a new filtered word we have to restart bot?

JasperGray: oh true. We don't have some time-to-live or reload filter cache command here yet

PR #48, opened 2020-05-05
Added Filter Commands

iShauny:


Floofbot/Services/Repository/Models/FilteredWord.cs
```diff
@@ -5,7 +5,7 @@ namespace Floofbot.Services.Repository.Models
     public partial class FilteredWord
     {
         [Key]
-        public ulong Id { get; set; }
+        public int Id { get; set; }
```
JasperGray, line 8: why downgrade to an int?

Floofbot/Modules/Filter.cs
```diff
@@ -0,0 +1,236 @@
...
+                }
+            }
+        }
+        [Summary("Lists all filtered words")]
+        [Command("list")]
+        public async Task ListFilteredWords()
+        {
+            List<FilteredWord> filteredWords = _floofDb.FilteredWords.AsQueryable()
+                .Where(x => x.ServerId == Context.Guild.Id)
+                .OrderBy(x => x.Id)
```
JasperGray, line 188: Sanity check: we're ordering by time of insertion into the DB rather than alphabetical order here?

Floofbot/Modules/Filter.cs
```diff
@@ -0,0 +1,236 @@
...
+        [Command("toggle")]
+        [Summary("Toggles the word filter")]
+        public async Task Toggle()
+        {
+            await Context.Channel.SendMessageAsync("", false, new EmbedBuilder { Description = $"💾 Usage: `filter toggle channel/server`", Color = EMBED_COLOR }.Build());
+        }
+
+        [Command("toggle")]
+        [Summary("Toggles the word filter")]
+        public async Task Toggle([Summary("channel/server")]string toggleType)
```
JasperGray, line 58: would be clearer here if the parameter summary was "either one of 'channel' or 'server'", and the method summary was "Toggles the server/channel-level word filter"

Floofbot/Modules/Filter.cs
```diff
@@ -0,0 +1,236 @@
...
+                        {
+                            ChannelId = Context.Channel.Id,
+                            ServerId = Context.Guild.Id
+                        });
+                        _floofDb.SaveChanges();
+                        channelInDatabase = true;
+                    }
+                    else
+                    {
+                        FilterChannelWhitelist channelConfig = _floofDb.FilterChannelWhitelists.Find(Context.Channel.Id);
```
JasperGray, line 101: Is this the same as channelData?

Floofbot/Modules/Filter.cs
```diff
@@ -0,0 +1,236 @@
...
+                    IsOn = false
+                });
+                _floofDb.SaveChanges();
+            }
+        }
+        private bool CheckWordEntryExists(string word, SocketGuild guild)
+        {
+            // checks if a word exists in the filter db
+            bool wordEntry = _floofDb.FilteredWords.AsQueryable().Where(w => w.ServerId == guild.Id).Where(w => w.Word == word).Any();
+            return (wordEntry == true) ? true : false;
```
JasperGray, line 47: We can just return `wordEntry` here

Floofbot/Services/Repository/Models/FilteredWord.cs
iShauny: sqlite does not support ordering by ulong

Floofbot/Modules/Filter.cs
iShauny: yes, i believe thats how the current filter works too

Floofbot/Modules/Filter.cs
iShauny: basically

Floofbot/Modules/Filter.cs
```diff
@@ -0,0 +1,235 @@
...
+        }
+        [Command("toggle")]
+        [Summary("Toggles the server/channel-level word filter")]
+        public async Task Toggle()
+        {
+            await Context.Channel.SendMessageAsync("", false, new EmbedBuilder { Description = $"💾 Usage: `filter toggle channel/server`", Color = EMBED_COLOR }.Build());
+        }
+
+        [Command("toggle")]
+        [Summary("Toggles the word filter")]
```
JasperGray, line 57: So if there's two summaries here (this toggle, and for the toggle method above), how is it going to look for the help command?

iShauny: I think it causes an issue at the moment, will probably need to update the help command to ignore duplicates

iShauny: Note: this already happens with the tag add command

PR #50, opened 2020-05-06
Finished Filter

iShauny:


Floofbot/Services/EventLoggerService.cs
```diff
@@ -56,7 +64,34 @@ public bool IsToggled(IGuild guild)
 
             return ServerConfig.IsOn;
         }
-
+        private async Task HandleBadMessage(SocketUser user, SocketMessage msg)
+        {
+            await msg.DeleteAsync();
+            var botMsg = await msg.Channel.SendMessageAsync($"{user.Mention} watch your language! You've said a bad word!");
+            Thread.Sleep(5000);
```
JasperGray, line 71: Judging by the linked SO question (and previous code in our codebase) I think we should go with `await Task.Delay(5000)` instead of `Thread.Sleep(5000)`, since the former goes along with the async-await architecture that we already have. 
https://stackoverflow.com/questions/20082221/when-to-use-task-delay-when-to-use-thread-sleep

Other than this, lgtm.

Floofbot/Services/EventLoggerService.cs
```diff
@@ -608,7 +657,35 @@ class WordFilterService
             List<FilteredWord> _filteredWords;
             DateTime _lastRefreshedTime;
 
-            bool hasFilteredWord(FloofDataContext floofDb, string messageContent, ulong serverId, ulong channelId)
+            public bool hasFilteredWord(FloofDataContext floofDb, string messageContent, ulong serverId) // names
```
JasperGray, line 660: It's nice that the comment's there, otherwise it'd be much harder to tell when to call which overload.

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -64,6 +64,9 @@ public async Task OnReactionAdded(Cacheable<IUserMessage, ulong> message, ISocke
         {
             var msg = message.Value as IUserMessage;
 
+            if (reaction.User.Value.IsBot)
```
JasperGray, line 67: Is this to prevent bots from calling this event handler?

iShauny: yes, the bot can trigger its own event handler

PR #51, opened 2020-05-06
Modified warning table.  Consolidated migrations

bealsbe:


Floofbot/Modules/Administration.cs
```diff
@@ -106,6 +106,11 @@ public class Administration : ModuleBase<SocketCommandContext>
                 return;
             }
 
+            if(reason.Length > 500) {
```
JasperGray, line 109: I guess you'll want to warn mods of this ahead of time around time of deployment.

Floofbot/Migrations/20200506001332_InitialCreate.cs
```diff
@@ -0,0 +1,164 @@
...
+                    Channel = table.Column<ulong>(nullable: false),
+                    IsOn = table.Column<bool>(nullable: false)
+                },
+                constraints: table =>
+                {
+                    table.PrimaryKey("PK_NicknameAlertConfigs", x => x.ServerId);
+                });
+
+            migrationBuilder.CreateTable(
+                name: "Tags",
```
JasperGray, line 101: lol this is nice.
what I could do in the future is add some code to slowly migrate the old tag format to the new tag format, since 
`${old tagname} = ${new tagname}:${serverId}` anyway.

Floofbot/Modules/Administration.cs
bealsbe: agreed

I don't think we have had a warning that was over this in the past

Floofbot/Migrations/20200506001332_InitialCreate.cs
bealsbe: I was thinking about doing some sublime magic on an exported CSV

PR #52, opened 2020-05-06
Refactor to get better summaries for help command

JasperGray:


Floofbot/Modules/TagCommands.cs
```diff
@@ -133,39 +127,39 @@ public async Task ListTags([Remainder] string content = null)
...
-        [Summary("Removes a tag")]
+        [Summary("Removes a tag from the server")]
         [RequireUserPermission(GuildPermission.ManageMessages)]
-        public async Task Remove([Summary("Tag name")] string tag)
+        public async Task Remove([Summary("Tag name")] string tag = null)
         {
-            string tagName = $"{tag.ToLower()}:{Context.Guild.Id}";
-            Tag tagToRemove = _floofDb.Tags.FirstOrDefault(x => x.TagName == tagName);
-            if (tagToRemove != null)
+            if (!string.IsNullOrEmpty(tag))
```
JasperGray, line 134: I basically wrapped the method in a null/empty check here.
The diff is not displaying the change efficiently

Floofbot/Modules/Logging.cs
```diff
@@ -26,7 +27,7 @@ public LoggerCommands(FloofDataContext floofDB)
                 _floofDB = floofDB;
             }
 
-            protected void CheckServer(ulong server)
+            private void CheckServer(ulong server)
```
bealsbe, line 30: What is the reason for changing these to private?

JasperGray: Usually, you want to set things to the lowest privilege. The logic I use is as follows:
If I give the code a quick glance and I see private, I know it's only being called within this class. If I see protected, I'll have to start worrying about where else it's being called.
If I have to give more access later, then I can do it later.

PR #56, opened 2020-05-06
Add 8ball responses to bot highlights that end with "?"

CryptidVulpes:
This how the change looks in action:

example

### DIFF
--- a/Floofbot/Handlers/CommandHandler.cs
+++ b/Floofbot/Handlers/CommandHandler.cs
@@ -61,13 +61,12 @@ namespace Floofbot.Handlers
 
             if (msg.HasStringPrefix(prefix, ref argPos) || msg.HasMentionPrefix(_client.CurrentUser, ref argPos))
             {
-                if(!msg.HasStringPrefix(prefix, ref argPos)) {
-                    string response = Floofbot.Modules.Fun.RespondEightBall(msg.ToString());
-                    if(response != "") {
-                        await context.Channel.SendMessageAsync(response);
-                    }
-                }
                 var result = await _commands.ExecuteAsync(context, argPos, _services);
+
+                if (result.Error == CommandError.UnknownCommand && msg.HasMentionPrefix(_client.CurrentUser, ref argPos) && msg.ToString().EndsWith("?")) {
+                    string response = Floofbot.Modules.Fun.QuestionEightBall(msg.ToString());
+                    await context.Channel.SendMessageAsync(response);
+                }
 
                 if (!result.IsSuccess && result.Error != CommandError.UnknownCommand)
                 {
--- a/Floofbot/Handlers/CommandHandler.cs
+++ b/Floofbot/Handlers/CommandHandler.cs
@@ -63,14 +63,14 @@ namespace Floofbot.Handlers
             {
                 var result = await _commands.ExecuteAsync(context, argPos, _services);
 
-                if (result.Error == CommandError.UnknownCommand && msg.HasMentionPrefix(_client.CurrentUser, ref argPos) && msg.ToString().EndsWith("?")) {
-                    string response = Floofbot.Modules.Fun.QuestionEightBall(msg.ToString());
-                    await context.Channel.SendMessageAsync(response);
-                }
-
                 if (!result.IsSuccess && result.Error != CommandError.UnknownCommand)
                 {
                     Log.Error(result.ErrorReason);
+                }
+
+                if (result.Error == CommandError.UnknownCommand && msg.HasMentionPrefix(_client.CurrentUser, ref argPos) && msg.ToString().EndsWith("?")) {
+                    string response = Floofbot.Modules.Fun.QuestionEightBall(msg.ToString());
+                    await context.Channel.SendMessageAsync(response);
                 }
             }
             else if (msg.Source == MessageSource.User)
--- a/Floofbot/Handlers/CommandHandler.cs
+++ b/Floofbot/Handlers/CommandHandler.cs
@@ -61,6 +61,12 @@ namespace Floofbot.Handlers
 
             if (msg.HasStringPrefix(prefix, ref argPos) || msg.HasMentionPrefix(_client.CurrentUser, ref argPos))
             {
+                if(!msg.HasStringPrefix(prefix, ref argPos)) {
+                    string response = Floofbot.Modules.Fun.RespondEightBall(msg.ToString());
+                    if(response != "") {
+                        await context.Channel.SendMessageAsync(response);
+                    }
+                }
                 var result = await _commands.ExecuteAsync(context, argPos, _services);
 
                 if (!result.IsSuccess && result.Error != CommandError.UnknownCommand)
--- a/Floofbot/Modules/Fun.cs
+++ b/Floofbot/Modules/Fun.cs
@@ -24,37 +24,9 @@ namespace Floofbot.Modules
             ".jpg", ".gif", ".png"
         };
 
-        public static string EightBallResponse() {
-            var responses = new List<string> {
-                 "As I see it, yes.",
-                 "Ask again later.",
-                 "Better not tell you now.",
-                 "Cannot predict now.",
-                 "Concentrate and ask again.",
-                 "Don’t count on it.",
-                 "It is certain.",
-                 "It is decidedly so.",
-                 "Most likely.",
-                 "My reply is no.",
-                 "My sources say no.",
-                 "Outlook not so good.",
-                 "Outlook good.",
-                 "Reply hazy, try again.",
-                 "Signs point to yes.",
-                 "Very doubtful.",
-                 "Without a doubt.",
-                 "Yes.",
-                 "Yes – definitely.",
-                 "You may rely on it."
-            };
-            Random random = new Random();
-            int randomNumber = random.Next(responses.Count);
-            return responses[randomNumber];
-        }
-
         public static string QuestionEightBall(string message)
         {
-            string response = EightBallResponse();
+            string response = EightBall.Response();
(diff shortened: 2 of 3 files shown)


Floofbot/Handlers/CommandHandler.cs
```diff
@@ -61,6 +61,12 @@ private async Task HandleCommandAsync(SocketMessage s)
 
             if (msg.HasStringPrefix(prefix, ref argPos) || msg.HasMentionPrefix(_client.CurrentUser, ref argPos))
             {
+                if(!msg.HasStringPrefix(prefix, ref argPos)) {
```
JasperGray, line 64: I think this should be `if msg.HasMentionPrefix(_client.CurrentUser, ref argPos)` instead to make it clearer that we're targeting mentions only.
Also, we might need to distinguish between responding to a question and running a command. 
For example, what if the command is `@Floofbot tag list example?`

Floofbot/Modules/Fun.cs
```diff
@@ -24,10 +24,7 @@ public class Fun : ModuleBase<SocketCommandContext>
             ".jpg", ".gif", ".png"
         };
 
-        [Command("8ball")]
-        [Summary("Ask the Magic 8-Ball a question")]
-        public async Task AskEightBall([Summary("question")][Remainder] string question)
-        {
+        public static string EightBallResponse() {
```
JasperGray, line 27: Since the 8-ball is being called at multiple places, we might want to place a new 8-ball class under Modules/Helpers and initialize the eight ball whenever we need it, instead.

Floofbot/Modules/Fun.cs
```diff
@@ -52,10 +49,33 @@ public async Task AskEightBall([Summary("question")][Remainder] string question)
...
+
+        [Command("8ball")]
+        [Summary("Ask the Magic 8-Ball a question")]
+        public async Task AskEightBall([Summary("question")][Remainder] string question)
+        {
             EmbedBuilder builder = new EmbedBuilder();
             builder.Title = "Magic 8 Ball";
             builder.AddField("Question", question);
-            builder.AddField("Answer", responses[randomNumber]);
+            builder.AddField("Answer", EightBallResponse());
```
JasperGray, line 78: `new EightBall().Response` would work better here once the class has been made.
Look into C# properties since those are used instead of Java-style getters and setters https://www.dotnetperls.com/property

Floofbot/Modules/Fun.cs
```diff
@@ -52,10 +49,33 @@ public async Task AskEightBall([Summary("question")][Remainder] string question)
             };
             Random random = new Random();
             int randomNumber = random.Next(responses.Count);
+            return responses[randomNumber];
+        }
+
+        public static string RespondEightBall(string question)
```
JasperGray, line 55: This method name is too close to the one above. It might become clearer what to do once EightBall moves into its own class.

Floofbot/Modules/Fun.cs
```diff
@@ -52,10 +49,33 @@ public async Task AskEightBall([Summary("question")][Remainder] string question)
...
+        {
+            Regex questionRe = new Regex(@"^.*\?$");
+            if(questionRe.IsMatch(question)) {
```
JasperGray, line 58: `if (question.EndsWith("?"))`

Floofbot/Modules/Fun.cs
```diff
@@ -52,10 +49,33 @@ public async Task AskEightBall([Summary("question")][Remainder] string question)
...
+                string response = Floofbot.Modules.Fun.EightBallResponse();
+                if (!string.IsNullOrEmpty(response))
+                {
+                    string reply = $"> {question}";
+                    reply += "\n";
+                    reply += $"{response}";
+                    return reply;
+                }
+            }
+            return "";
```
JasperGray, line 68: you can return `string.Empty` here just to make it clear that you didn't forget to fill in anything between the quotes

Floofbot/Handlers/CommandHandler.cs
```diff
@@ -61,6 +61,12 @@ private async Task HandleCommandAsync(SocketMessage s)
...
+                    string response = Floofbot.Modules.Fun.RespondEightBall(msg.ToString());
+                    if(response != "") {
```
JasperGray, line 66: optional: `if (!string.isNullOrEmpty(response))` for readability

Floofbot/Modules/Fun.cs
```diff
@@ -52,10 +49,33 @@ public async Task AskEightBall([Summary("question")][Remainder] string question)
...
+        {
+            Regex questionRe = new Regex(@"^.*\?$");
+            if(questionRe.IsMatch(question)) {
+                string response = Floofbot.Modules.Fun.EightBallResponse();
+                if (!string.IsNullOrEmpty(response))
+                {
+                    string reply = $"> {question}";
+                    reply += "\n";
+                    reply += $"{response}";
+                    return reply;
```
JasperGray, line 65: I think you can just `return $"> {question}\n{response}"` directly, without even having to create the new variable, unless I'm missing something?

CryptidVulpes: You're right, but I thought this would be readable, as the internal representation is multiline as well.

Floofbot/Modules/Fun.cs
CryptidVulpes: I'm not sure how C# internals go, but I wouldn't like to instantiate a new object every time we send an EightBall response. Sounds like a way to lose memory and/or over-stress the garbage collector.

Floofbot/Handlers/CommandHandler.cs
CryptidVulpes: > I think this should be if msg.HasMentionPrefix(_client.CurrentUser, ref argPos) instead to make it clearer that we're targeting mentions only.
True, will fix

> Also, we might need to distinguish between responding to a question and running a command.
> For example, what if the command is @Floofbot tag list example?

You're right. I'm not sure how to fix this though...

Floofbot/Modules/Fun.cs
CryptidVulpes: What about `QuestionEightBall()` ?

Floofbot/Handlers/CommandHandler.cs
CryptidVulpes: I'll check for `result.Error == CommandError.UnknownCommand` and fire the `QuestionEightBall` only if that happens. Should be robust enough?

Floofbot/Modules/Fun.cs
JasperGray: It actually confuses me more because I have to mentally concat them, but maybe that's just me :3

CryptidVulpes: I'll make it a oneliner ;)

Floofbot/Modules/Fun.cs
JasperGray: True. You can make it a class with just static members then to keep things organized, since otherwise the bot is calling directly into this module sometimes, which should only be called by valid commands.

Floofbot/Modules/Fun.cs
CryptidVulpes: Created a static helper class

Floofbot/Modules/Fun.cs
CryptidVulpes: Created a static helper class

CryptidVulpes:
Removed the `isNullOrEmpty` checks, since the helper function should always return something (or raise an error)

Floofbot/Handlers/CommandHandler.cs
```diff
@@ -63,6 +63,11 @@ private async Task HandleCommandAsync(SocketMessage s)
             {
                 var result = await _commands.ExecuteAsync(context, argPos, _services);
 
+                if (result.Error == CommandError.UnknownCommand && msg.HasMentionPrefix(_client.CurrentUser, ref argPos) && msg.ToString().EndsWith("?")) {
```
CryptidVulpes, line 66: I moved the test for question to the handler, since I think it should not make extra calls to modules that might not do anything.

Floofbot/Handlers/CommandHandler.cs
```diff
@@ -63,6 +63,11 @@ private async Task HandleCommandAsync(SocketMessage s)
...
+                    string response = Floofbot.Modules.Fun.QuestionEightBall(msg.ToString());
+                    await context.Channel.SendMessageAsync(response);
```
JasperGray, line 68: Nice update here.
Sanity check: do we want to return after this, or run the rest of the code in this method?

Floofbot/Modules/Helpers/EightBall.cs
```diff
@@ -0,0 +1,36 @@
+using System;
+using System.Collections.Generic;
+
+namespace Floofbot.Modules.Helpers
+{
+    public static class EightBall
+    {
+        public static string Response() {
+            var responses = new List<string> {
```
JasperGray, line 9: We can make responses a `private static readonly List<string>` variable within this class now!

JasperGray: maybe rename this to GetRandomResponse()

Floofbot/Handlers/CommandHandler.cs
CryptidVulpes: You're absolutely right. Does it sound better to move this `if` block after the `Log.Error` block and `await` there?

CryptidVulpes: I moved the EightBall response to after the error logging.

PR #59, opened 2020-05-06
Refactor help module

JasperGray:


Floofbot/Modules/Help.cs
```diff
@@ -57,73 +64,57 @@ public async Task HelpCommand()
...
-            // put available module names into a list instead of their object
-            foreach (ModuleInfo module in _commands.Modules.ToList())
-                moduleNames.Add(module.Name.ToLower());
-
-            if (!moduleNames.Contains(requestedModule.ToLower())) // no such cmd
+            List<string> moduleNames = _commandService.Modules.Select(x => x.Name.ToLower()).ToList();
+            if (!moduleNames.Contains(requestedModule.ToLower()))
             {
-                await Context.Channel.SendMessageAsync("Unable to find that command");
+                await Context.Channel.SendMessageAsync($"Unable to find the '{requestedModule}' module");
```
iShauny, line 79: I would put the word command instead of module as it will be users using this and they wont be aware of the concept of 'modules'

JasperGray: Fixed. I also added an explanation for modules in the command summary because otherwise the user might start trying things like `.help ping` when it should be `.help utilities`.

PR #64, opened 2020-05-10
Add tag requireadmin command and add/remove permission checks

JasperGray:


Floofbot/Modules/TagCommands.cs
```diff
@@ -33,13 +34,75 @@ public TagCommands(FloofDataContext floofDb)
...
+                        ServerId = Context.Guild.Id,
+                        TagUpdateRequiresAdmin = parsedRequireAdmin,
+                    });
+                }
+                else
+                {
+                    config.TagUpdateRequiresAdmin = parsedRequireAdmin;
+                }
+                await _floofDb.SaveChangesAsync();
+                string message = $"Adding/removing tags now{(parsedRequireAdmin ? " requires " : " does not require ")}admin permission.";
```
iShauny, line 67: minor thing but these lines are a bit too long, perhaps split them up

Floofbot/Modules/TagCommands.cs
```diff
@@ -33,13 +34,75 @@ public TagCommands(FloofDataContext floofDb)
...
+                await Context.Channel.SendMessageAsync(message);
+            }
+            catch (DbUpdateException e)
+            {
+                string message = $"Unable to make adding/removing tags{(parsedRequireAdmin ? " " : " not ")}require admin permission.";
+                await Context.Channel.SendMessageAsync(message);
+                Log.Error(e.ToString());
```
iShauny, line 74: I think it would be better to give more context to these exceptions eg "Error when configuring tag admin permissions" or something

Floofbot/Modules/TagCommands.cs
```diff
@@ -58,10 +121,9 @@ public TagCommands(FloofDataContext floofDb)
                     _floofDb.SaveChanges();
                     await SendEmbed(CreateDescriptionEmbed($"💾 Added Tag `{tag}`"));
                 }
-                catch (DbUpdateException e)
+                catch (DbUpdateException)
```
iShauny, line 124: I kind of dont like this, other exception errors could occur regarding the database that are not just entries already existing

Perhaps checking if it exists first is better?

PR #65, opened 2020-05-10
Add choice command

JasperGray:


Floofbot/Modules/Fun.cs
```diff
@@ -170,6 +172,23 @@ public async Task RequestBirb()
             }
         }
 
+        [Command("choice")]
+        [Summary("Chooses an option")]
```
iShauny, line 176: I would expand this summary as its not very clear

maybe
"Choose one item from a list of possible options" or something

Floofbot/Modules/Fun.cs
```diff
@@ -170,6 +172,23 @@ public async Task RequestBirb()
...
+            {
+                string[] splitChoices = choices.Split(";", StringSplitOptions.RemoveEmptyEntries)
+                    .Select(choice => choice.Trim()).ToArray();
+                if (splitChoices.Length != 0)
+                {
+                    await Context.Channel.SendMessageAsync(splitChoices[rand.Next(splitChoices.Length)]);
+                    return;
+                }
+            }
+            await Context.Channel.SendMessageAsync("Not enough options were provided.");
```
iShauny, line 189: I would send something explaining how to use the delimiter, maybe provide an example

PR #66, opened 2020-05-10
Refactor logging and allow setchannel to take fewer args

JasperGray:


Floofbot/Modules/Logging.cs
```diff
@@ -20,79 +20,123 @@ public class Logging
...
+                    {
+                        await Context.Channel.SendMessageAsync($"Unable to set {channelType} to <#{channel.Id}>");
+                    }
+                }
+                else
+                {
+                    EmbedBuilder builder = new EmbedBuilder()
+                    {
+                        Description = $"💾 Accepted channel types: ```{string.Join(", ", CHANNEL_TYPES)}```",
+                        Color = Color.Magenta
```
iShauny, line 114: Should colour be a constant?

JasperGray: For now, yeah
If we're randomizing the colour, I'd rather have a discussion on which parts should be randomized first.

PR #67, opened 2020-05-10
Move EightBall question

CryptidVulpes:
The behavior is now moved to `EventLoggerService.OnMessage` so it will be run in an `async` context

### DIFF
--- a/Floofbot/Services/EventLoggerService.cs
+++ b/Floofbot/Services/EventLoggerService.cs
@@ -1,4 +1,5 @@
 ﻿using Discord;
+using Discord.Commands;
 using Discord.WebSocket;
 using Floofbot.Services.Repository;
 using Floofbot.Services.Repository.Models;
@@ -73,6 +74,8 @@ namespace Floofbot.Services
         }
         public Task OnMessage(SocketMessage msg)
         {
+            var userMsg = msg as SocketUserMessage;
+            int argPos = 0;
             var _ = Task.Run(async () =>
             {
                 try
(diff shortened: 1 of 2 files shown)


Floofbot/Services/EventLoggerService.cs
```diff
@@ -73,16 +74,24 @@ private async Task HandleBadMessage(SocketUser user, SocketMessage msg)
...
                 try
                 {
                     if (msg.Author.IsBot)
                         return;
                     var channel = msg.Channel as ITextChannel;
+                    string content = msg.Content;
                     bool hasBadWord = _wordFilterService.hasFilteredWord(new FloofDataContext(), msg.Content, channel.Guild.Id, msg.Channel.Id);
                     if (hasBadWord)
                         await HandleBadMessage(msg.Author, msg);
+                    if (userMsg.HasMentionPrefix(_client.CurrentUser, ref argPos) && content.EndsWith("?"))
```
JasperGray, line 90: shouldn't we use else if here, because we don't want the 8ball to answer the question?

Floofbot/Services/EventLoggerService.cs
```diff
@@ -73,16 +74,24 @@ private async Task HandleBadMessage(SocketUser user, SocketMessage msg)
         }
         public Task OnMessage(SocketMessage msg)
         {
+            var userMsg = msg as SocketUserMessage;
```
JasperGray, line 77: I think we'll have to do a null check later on if this returns null.
Can we use an explicit cast `(SocketUserMessage)msg` instead, if this cast will always succeed?

Floofbot/Services/EventLoggerService.cs
CryptidVulpes: Doesn't the `await HandleBadMessage(msg.Author, msg);` return from the function anyway?

Floofbot/Services/EventLoggerService.cs
CryptidVulpes: I'll test for the `null`. The original function in `CommandHandler.cs` tests it as well.

Floofbot/Services/EventLoggerService.cs
CryptidVulpes: Added the `else if` anyway

JasperGray: await is a keyword used with calling an async function. If we want to return we'd explicitly have to use the return keyword, e.g. `return await ...`.

PR #68, opened 2020-05-11
Use new schema for tags

JasperGray:


Floofbot/Modules/TagCommands.cs
```diff
@@ -77,6 +77,34 @@ public async Task RequireAdmin([Summary("True/False")] string requireAdmin = "")
             }
         }
 
+        [Command("updatetagnames")]
+        [Summary("Update tag names to new schema")]
+        [RequireUserPermission(GuildPermission.Administrator)]
+        public async Task UpdateTagNames(string dummyParam="")
```
JasperGray, line 83: @ Reviewers: Any idea what to do here instead of dummyParam?
If I take off this parameter, even if I set the command's priority to 0, the general `.tag` command gets called and says it can't find a tag with name `updatetagnames`.

iShauny: If this is only a one time command, why not just not put it under the tag group? make it a command of its own

Floofbot/Modules/TagCommands.cs
```diff
@@ -105,34 +105,39 @@ private bool UserHasTagUpdatePermissions(IGuildUser user)
...
+            if (!string.IsNullOrEmpty(tagName) && !string.IsNullOrEmpty(content))
             {
-                Regex rgx = new Regex("[^a-zA-Z0-9 -]");
-                tag = rgx.Replace(tag, "").ToLower();
-                string tagName = $"{tag.ToString()}:{Context.Guild.Id}";
+                Regex rgx = new Regex("[^a-zA-Z0-9-]");
+                string processedTagName = rgx.Replace(tagName, "").ToLower();
+                if (string.IsNullOrEmpty(processedTagName))
+                {
+                    await SendEmbed(CreateDescriptionEmbed($"💾 Invalid Tag name. Tag must contain characters within [A-Za-z0-9-]."));
```
iShauny, line 114: I'm not sure because of githubs way of displaying it, but the line seems a bit long and perhaps breaking it up is good

JasperGray: Fixed

PR #69, opened 2020-05-11
Move RandomResponseGenerator invokation to event handler

CryptidVulpes:


Floofbot/Services/EventLoggerService.cs
```diff
@@ -92,6 +92,14 @@ public Task OnMessage(SocketMessage msg)
                         string eightBallResponse = Floofbot.Modules.Helpers.EightBall.GetRandomResponse();
                         await channel.SendMessageAsync($"> {content}\n{eightBallResponse}");
                     }
+                    else if (userMsg != null)
```
iShauny, line 95: This if statement is kinda not neccessary. You already check in your previous 8ball part if the message is null, but even then, you can do a null check at the start of the task anyway

CryptidVulpes: I don't want to check for `null` in the beginning, because I would either skip the bad word test, or I would have to anyway test for `null` in these `if`-clauses. The only other way I can think of right now is to test like this:

```csharp
if (hasBadWord)
    await HandleBadMessage(msg.Author, msg);
else if (userMsg == null)
    return;
else if (userMsg.HasMentionPrefix(_client.CurrentUser, ref argPos) && content.EndsWith("?"))
{
    string eightBallResponse = Floofbot.Modules.Helpers.EightBall.GetRandomResponse();
    await channel.SendMessageAsync($"> {content}\n{eightBallResponse}");
}
else
{
    string randomResponse = RandomResponseGenerator.GenerateResponse(userMsg);
    if (!string.IsNullOrEmpty(randomResponse))
    {
        await channel.SendMessageAsync(randomResponse);
    }
}
```

I think the current proposal is more readable than that 🤔

iShauny: The filter is pointless if the usermessage is null anyway, so the null check should happen before the filter check.

JasperGray: SocketUserMessage extends SocketMessage, so the `as` cast will always succeed.
Therefore, we do not need to check if `userMsg` is null.
If `userMsg` is actually null, then `msg` would have to be null, causing an exception at the `if (msg.Author.IsBot)` check.
Therefore if we want to be safe, then it's better to just change that condition to `if (msg == null || msg.Author.IsBot)`

iShauny: agree with above

CryptidVulpes: Very true. Will do that change.

PR #70, opened 2020-05-11
Add Joined Server timestamp and Time At Server interval to UserLeft

CryptidVulpes:


Floofbot/Services/EventLoggerService.cs
```diff
@@ -389,11 +389,18 @@ public Task UserLeft(IGuildUser user)
...
-                         .WithDescription($"{user.Mention} | ``{user.Id}``")
-                         .WithFooter($"user_leave user_joinlog {user.Id}")
+                         .WithDescription($"{user.Mention} | ``{user.Id}``");
+                    if (user.JoinedAt != null)
+                    {
+                        DateTimeOffset userJoined = ((DateTimeOffset)user.JoinedAt);
+                        TimeSpan interval = DateTime.UtcNow - userJoined.DateTime;
+                        string day_word = interval.Days == 1 ? "day" : "days";
+                        embed.AddField("Joined Server", userJoined.ToString("ddd, dd MMM yyyy"), true);
+                        embed.AddField("Time at Server", $"{interval.Days} {day_word}", true);
```
iShauny, line 401: Overall lgtm, however, wouldnt it be better to have it show years/months too? having, say, 760 days in the server is not very readable. Not PR blocking though just food for thought

Floofbot/Services/EventLoggerService.cs
```diff
@@ -389,11 +389,18 @@ public Task UserLeft(IGuildUser user)
...
-
                     embed.WithTitle($"❌ User Left | {user.Username}#{user.Discriminator}")
                          .WithColor(Color.Red)
-                         .WithDescription($"{user.Mention} | ``{user.Id}``")
-                         .WithFooter($"user_leave user_joinlog {user.Id}")
+                         .WithDescription($"{user.Mention} | ``{user.Id}``");
+                    if (user.JoinedAt != null)
+                    {
+                        DateTimeOffset userJoined = ((DateTimeOffset)user.JoinedAt);
+                        TimeSpan interval = DateTime.UtcNow - userJoined.DateTime;
```
JasperGray, line 398: We're using UtcNow here?
I thought we were going to use the local server time which would just be DateTime.Now?

Floofbot/Services/EventLoggerService.cs
CryptidVulpes: Hmm yes. You are right. Unfortunately `TimeSpan` object doesn't count years or months, so some arithmetic needs to be done here to implement that.

JasperGray, line 401: "Time at Server" sounds like "current local time on the server". I wonder if we can come up with another description for this field, such as "Amount of Time in Server", or if we're sticking to just days, "# Days in Server"?
I do think using yrs/months would help too, although I think it depends on what we're doing.
For example, if we're going to scrape over all of the leave times then using just days is better.

Floofbot/Services/EventLoggerService.cs
CryptidVulpes: The value returned from the API is UTC, so I need to use UtcNow to get an accurate interval.

Floofbot/Services/EventLoggerService.cs
CryptidVulpes: From the context and valua, it should be easily seen it's not the local time of the server. Though maybe "Member for" or something like that could be better.

PR #71, opened 2020-05-12
Add a gitattributes file

JasperGray:


JasperGray:
Closing -- we already have a .gitattributes file?

PR #73, opened 2020-05-12
Added No Action Option to Nickname Alert 

iShauny:


Floofbot/Services/NicknameAlertService.cs
```diff
@@ -45,7 +46,8 @@ public async Task HandleBadNickname(SocketGuildUser badUser, IGuild guild)
                 .WithDescription($"{REMOVE_NICKNAME_EMOJI.Name}: Remove Nickname\n" +
                 $"{WARN_EMOJI.Name}: Warn\n" +
                 $"{KICK_EMOJI.Name}: Kick\n" +
-                $"{BAN_EMOJI.Name}: Ban")
+                $"{BAN_EMOJI.Name}: Ban\n" +
+                $"{NO_ACTION_EMOJI}: No Action")
```
JasperGray, line 50: I think we need to call .Name here

Floofbot/Services/NicknameAlertService.cs
```diff
@@ -166,6 +169,11 @@ public async Task OnReactionAdded(Cacheable<IUserMessage, ulong> message, ISocke
                     }
                     alertMessageIdsDic.Remove(msg.Id);
                 }
+                else if (reaction.Emote.Name.Equals(NO_ACTION_EMOJI.Name))
```
JasperGray, line 172: Actually, for the REMOVE option should we change the message to "I removed ..." rather than "I changed ..."?

PR #77, opened 2020-06-04
Add tag update command

JasperGray:


JasperGray:
> does this command check if admin only tag management is enabled and if they are admin?

Yes because UserHasTagUpdatePermissions is called

PR #78, opened 2020-06-04
Create purge command

JasperGray:
Warning: if we're logging deletes for the purge command, then the delete messages will be rate-limited because the deletions are occurring too quickly.

### DIFF
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -1,11 +1,12 @@
-﻿using System;
-using Discord;
-using System.Threading.Tasks;
+﻿using Discord;
 using Discord.Commands;
-using System.Text.RegularExpressions;
+using Discord.WebSocket;
 using Floofbot.Services.Repository;
 using Floofbot.Services.Repository.Models;
+using System;
 using System.Linq;
+using System.Threading.Tasks;
+using System.Text.RegularExpressions;
 
 namespace Floofbot.Modules
 {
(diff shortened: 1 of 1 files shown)


Floofbot/Modules/Administration.cs
```diff
@@ -146,6 +148,45 @@ public class Administration : ModuleBase<SocketCommandContext>
...
+            foreach (ISocketMessageChannel channel in Context.Guild.TextChannels)
+            {
+                var asyncMessageCollections = channel.GetMessagesAsync(MESSAGES_TO_SCAN_PER_CHANNEL_ON_PURGE);
+                await foreach(var messageCollection in asyncMessageCollections)
+                {
+                    foreach (var message in messageCollection)
+                    {
+                        if (message.Author.Id == badUser.Id)
+                        {
+                            await channel.DeleteMessageAsync(message);
```
iShauny, line 175: could run this in a delayed async task to help with rate limits async task

Floofbot/Modules/Administration.cs
```diff
@@ -146,6 +148,45 @@ public class Administration : ModuleBase<SocketCommandContext>
...
 
+        [Command("purge")]
+        [Alias("p")]
+        [Summary("Deletes recent messages from a given user for all channels on the server")]
+        [RequireContext(ContextType.Guild)]
+        [RequireUserPermission(GuildPermission.BanMembers)]
+        public async Task PurgeUserMessages(
+            [Summary("user")] string user)
+        {
+            IUser badUser = resolveUser(user);
```
iShauny, line 159: will this command work after a user is banned?

Floofbot/Modules/Administration.cs
JasperGray: The part that needs to be delayed isn't exactly the deletion, but the logging of the messages

Floofbot/Modules/Administration.cs
JasperGray: All it has to do is resolve the parameter user, which is either the userid or the username#tag.
I'm thinking that if banning a user after they already left the server works, then this should work too.

Floofbot/Modules/Administration.cs
iShauny: delaying the deletion will delay the logging, though

Floofbot/Modules/Administration.cs
iShauny: it does not, though :( I think we wont be able to resolve this until we implement a user cache

PR #86, opened 2020-07-08
Added commands to forgive/unforgive warnings

iShauny:
+ Commands have been added that allow moderators to forgive and unforgive warnings/usernotes
+ Warnlog has been updated to show this

### DIFF
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -692,7 +692,7 @@ namespace Floofbot.Modules
             {
                 if (!_floofDB.UserNotes.AsQueryable().Where(w => w.UserId == uID && w.GuildId == Context.Guild.Id && w.Forgiven == true).Any()) // there are no user notes for this user in this guild
                 {
-                    await Context.Channel.SendMessageAsync("User has no warnings to unforgive!");
+                    await Context.Channel.SendMessageAsync("User has no user notes to unforgive!");
                     return;
                 }
                 warnings = _floofDB.UserNotes.AsQueryable()
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -10,6 +10,7 @@ using System.Text.RegularExpressions;
 using System.Collections.Generic;
 using Discord.Addons.Interactive;
 using Microsoft.EntityFrameworkCore;
+using Serilog;
 
 namespace Floofbot.Modules
 {
@@ -570,6 +571,7 @@ namespace Floofbot.Modules
                 builder.WithTitle($"Warnings for {user.Username}#{user.Discriminator}");
             if (warnings == null) // for some reason didnt recieve data from database
             {
+                Log.Error("Fatal error when trying to access warnings for the forgive user command!");
                 return;
             }
             if (type == "warning")
@@ -707,6 +709,7 @@ namespace Floofbot.Modules
                 builder.WithTitle($"Forgiven warnings for {user.Username}#{user.Discriminator}");
             if (warnings == null) // for some reason didnt recieve data from database
             {
+                Log.Error("Fatal error when trying to access warnings for the forgive user command!");
                 return;
             }
             if (type == "warning")
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -555,7 +555,7 @@ namespace Floofbot.Modules
             {
                 if (!_floofDB.UserNotes.AsQueryable().Where(w => w.UserId == uID && w.GuildId == Context.Guild.Id && w.Forgiven == false).Any()) // there are no user notes for this user in this guild
                 {
-                    await Context.Channel.SendMessageAsync("User has no warnings to forgive!");
+                    await Context.Channel.SendMessageAsync("User has no user notes to forgive!");
                     return;
                 }
                 warnings = _floofDB.UserNotes.AsQueryable()
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -464,7 +464,10 @@ namespace Floofbot.Modules
             builder.Color = ADMIN_COLOR;
             int warningCount = 0;
             int userNoteCount = 0;
-            builder.WithTitle($"Warnings for {badUser.Username}#{badUser.Discriminator}");
+            if (badUser == null) // no user, just id in database
+                builder.WithTitle($"Warnings for {user}");
+            else
+                builder.WithTitle($"Warnings for {badUser.Username}#{badUser.Discriminator}");
             if (formalWarnings.Count() != 0) // they have warnings
             {
                 builder.AddField(":warning: | Formal Warnings:", "\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_");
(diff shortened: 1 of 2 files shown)


Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +509,396 @@ public async Task warnlog([Summary("user")] string user)
...
+                    if (type == "warning") // forgive warning
+                    {
+                        if (!_floofDB.Warnings.AsQueryable().Where(w => w.UserId == Convert.ToUInt64(badUser) && w.GuildId == Context.Guild.Id && w.Forgiven == false).Any()) // there are no warnings for this user in this guild
+                        {
+                            await Context.Channel.SendMessageAsync("User has no warnings to forgive!");
+                            return;
+                        }
+                        warnings = _floofDB.Warnings.AsQueryable()
+                            .Where(u => u.UserId == Convert.ToUInt64(badUser) && u.GuildId == Context.Guild.Id && u.Forgiven == false)
+                            .OrderByDescending(x => x.DateAdded).Take(10);
```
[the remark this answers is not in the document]
iShauny: We will rarely ever have more than 10 warnings/user notes per user, but if we do, they will be so old that they can be ignored and forgotten, even in the warnlog

Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +509,396 @@ public async Task warnlog([Summary("user")] string user)
...
         }
 
+        [Command("forgive", RunMode = RunMode.Async)]
+        [Summary("Remove a user's warning or user notes")]
+        [RequireContext(ContextType.Guild)]
+        [RequireUserPermission(GuildPermission.BanMembers)]
+        public async Task forgiveUser([Summary("warning/usernote")] string type = "", [Summary("user")] string badUser = "")
+        {
+            IQueryable warnings = null;
+            if (string.IsNullOrEmpty(type) || string.IsNullOrEmpty(badUser) || (!type.ToLower().Equals("warning") && !type.ToLower().Equals("usernote"))) // invalid parameters
```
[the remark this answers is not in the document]
iShauny: can only be done if the help module is recoded in a way that supports groups better

Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +509,396 @@ public async Task warnlog([Summary("user")] string user)
...
+            // handle user input
+            if (type == "warning")
+            {
+                EmbedBuilder builder = new EmbedBuilder();
+                builder.Color = ADMIN_COLOR;
+                if (user == null) // no user, just id in database
+                    builder.WithTitle($"Warnings for {badUser}");
+                else
+                    builder.WithTitle($"Warnings for {user.Username}#{user.Discriminator}");
+                if (warnings == null) // for some reason didnt recieve data from database
```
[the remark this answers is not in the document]
iShauny: I added  fatal error log

Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +509,396 @@ public async Task warnlog([Summary("user")] string user)
...
+            }
+            IUser user = resolveUser(badUser);
+            if (user == null)
+            {
+                // user not in server, check if user is in database
+                if (Regex.IsMatch(badUser, @"\d{17,18}"))
+                {
+                    if (type == "warning") // forgive warning
+                    {
+                        if (!_floofDB.Warnings.AsQueryable().Where(w => w.UserId == Convert.ToUInt64(badUser) && w.GuildId == Context.Guild.Id && w.Forgiven == false).Any()) // there are no warnings for this user in this guild
```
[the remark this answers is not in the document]
iShauny: I've refactored the code to encapsulate similar functions

Floofbot/Modules/Administration.cs
```diff
@@ -464,7 +464,10 @@ public async Task warnlog([Summary("user")] string user)
             builder.Color = ADMIN_COLOR;
             int warningCount = 0;
             int userNoteCount = 0;
-            builder.WithTitle($"Warnings for {badUser.Username}#{badUser.Discriminator}");
+            if (badUser == null) // no user, just id in database
```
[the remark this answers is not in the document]
iShauny: refactored code

Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +510,280 @@ public async Task warnlog([Summary("user")] string user)
...
+                }
+                warnings = _floofDB.UserNotes.AsQueryable()
+                    .Where(u => u.UserId == uID && u.GuildId == Context.Guild.Id && u.Forgiven == false)
+                    .OrderByDescending(x => x.DateAdded).Take(10);
+            }
+
+            EmbedBuilder builder = new EmbedBuilder();
+            builder.Color = ADMIN_COLOR;
+            if (user == null) // no user, just id in database
+                builder.WithTitle($"Warnings for {badUser}");
```
CryptidVulpes, line 569: The bot will unfortunately now output `Warnings for` even if the moderator wanted to manipulate user notes.

CryptidVulpes: This same thing is in lines 574, 707 and 712

Floofbot/Modules/Administration.cs
```diff
@@ -473,7 +477,7 @@ public async Task warnlog([Summary("user")] string user)
                     if (warning.Forgiven)
                     {
                         IUser forgivenBy = resolveUser(warning.ForgivenBy.ToString());
-                        var forgivenByText = (forgivenBy != null) ? "" : $"(forgiven by {forgivenBy.Username}#{forgivenBy.Discriminator}";
+                        var forgivenByText = (forgivenBy == null) ? "" : $"(forgiven by {forgivenBy.Username}#{forgivenBy.Discriminator})";
```
CryptidVulpes, line 480: In the below functions, `forgivenBy` is always set when `Forgiven` is `true`. Is it actually possible for it to be `null`?

Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +510,280 @@ public async Task warnlog([Summary("user")] string user)
...
+                {
+                    builder.AddField($"**ID: {w.Id}** - {w.DateAdded.ToString("yyyy MMMM dd")} - {w.Moderator}", $"```{w.Reason}```");
+                }
+            }
+            await SendEmbed(builder.Build());
+            try
+            {
+                await ReplyAsync("Which would you like to forgive? Please specify the ID.");
+
+                SocketMessage response = await NextMessageAsync(true, true, TimeSpan.FromSeconds(10)); // wait for reply from source user in source channel for 2 seconds
```
CryptidVulpes, line 596: The earlier comment was probably lost, but here the code says 10 seconds, but your comment says 2 seconds. Which one did you want?

CryptidVulpes: Same thing is in the unforgive function.

Floofbot/Modules/Administration.cs
```diff
@@ -506,6 +510,280 @@ public async Task warnlog([Summary("user")] string user)
...
+            if (type == "warning") // forgive warning
+            {
+                if (!_floofDB.Warnings.AsQueryable().Where(w => w.UserId == uID && w.GuildId == Context.Guild.Id && w.Forgiven == false).Any()) // there are no warnings for this user in this guild
+                {
+                    await Context.Channel.SendMessageAsync("User has no warnings to forgive!");
+                    return;
+                }
+                warnings = _floofDB.Warnings.AsQueryable()
+                    .Where(u => u.UserId == uID && u.GuildId == Context.Guild.Id && u.Forgiven == false)
+                    .OrderByDescending(x => x.DateAdded).Take(10);
```
CryptidVulpes, line 552: Earlier comment was probably lost, but here you `Take(10)` which only returns the ten latest warnings/notes. Is this desired? What if the moderator wants to query for older ones and the user has more than ten?

This same thing is in lines 563, 689, and 700.

Floofbot/Modules/Administration.cs
iShauny: Added a ternary to fix this in refactoring

Floofbot/Modules/Administration.cs
iShauny: If the moderator that forgave the user is no longer in the server, the function will not be able to resolve the user ID, so the function will return a null result

Floofbot/Modules/Administration.cs
iShauny: Fixed

Floofbot/Modules/Administration.cs
iShauny: There will almost never be more than 10 warnings / user notes, however, if there are, we just want to ignore them as thats what was desired in the existing warnlog function and it doesnt seem to make sense to have more warnings in one than the other

PR #89, opened 2020-07-09
Created modmail command

iShauny:
It is hoped that in a future PR this can be refactored and the fixed server ID can be replaced with a system that supports multiple servers


Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+                Log.Error("Error when trying to toggle the modmail: " + ex);
+                return;
+            }
+        }
+        [Command("modrole")]
+        [Summary("OPTIONAL: A Role to Ping When ModMail is Received.")]
+        public async Task SetModRole(string roleName = null)
+        {
+            var ServerConfig = _floofDB.ModMails.Find(Context.Guild.Id);
+            if (roleName == null)
```
jkchen2, line 167: Invalid argument check should happen first

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+                        await Context.Channel.SendMessageAsync("Mod role set!");
+                        _floofDB.SaveChanges();
+                        return;
+                    }
+                    catch (Exception ex)
+                    {
+                        await Context.Channel.SendMessageAsync("An error occured: " + ex.Message);
+                        Log.Error("Error when trying to set the modmail mod role: " + ex);
+                        return;
+                    }
```
jkchen2, line 188: May want to chuck a `break` in here. Discord allows the existence of multiple roles with the same name. An optional refactor would be to filter the role list first, check to see if there are any duplicates, and raise an error based on that -- but that's pretty overkill for this project.

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+                {
+                    return;
+                }
+
+                if (serverConfig.IsEnabled == false || guild == null || channel == null) // disabled OR channel for mails not set
+                {
+                    return;
+                }
+
+                if (string.IsNullOrEmpty(content))
```
jkchen2, line 53: Content checking/command usage should probably be up top

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+                    ModRoleId = null
+                });
+                _floofDB.SaveChanges();
+            }
+        }
+
+        [Command("channel")]
+        [Summary("Sets the channel for the modmail notifications")]
+        public async Task Channel([Summary("Channel (eg #alerts)")]Discord.IChannel channel)
+        {
```
jkchen2, line 127: No checks for valid arguments?

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+            try
+            {
+                // get values
+                var serverConfig = _floofDb.ModMails.Find(RFURRY_SERVER_ID);
+                IGuild guild = Context.Client.GetGuild(RFURRY_SERVER_ID); // can return null
+                Discord.ITextChannel channel = await guild.GetTextChannelAsync((ulong)serverConfig.ChannelId); // can return null
+                IRole role = null;
+                if (serverConfig.ModRoleId != null)
+                {
+                    role = guild.GetRole((ulong)serverConfig.ModRoleId); // can return null
```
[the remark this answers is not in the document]
iShauny: The ulong cast is required because the database is stored as a nullable ulong (ulong?) and when using the GetRole function etc they cannot implicitly case ulong? to ulong. A get around is to change the DB so instead of using a nullable ulong it sets the value to 0. Mod role is optional and channel can be empty if it isnt configured. I don't think either solution matters though.

I moved the early return checking

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+
+        [Command("")]
+        public async Task sendModMail([Summary("Message Content")][Remainder] string content)
+        {
+            try
+            {
+                // get values
+                var serverConfig = _floofDb.ModMails.Find(RFURRY_SERVER_ID);
+                IGuild guild = Context.Client.GetGuild(RFURRY_SERVER_ID); // can return null
+                Discord.ITextChannel channel = await guild.GetTextChannelAsync((ulong)serverConfig.ChannelId); // can return null
```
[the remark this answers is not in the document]
iShauny: refer to previous comment

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+                    {
+                        Description = $"Usage: `modmail [message]`",
+                        Color = Color.Magenta
+                    };
+                    await Context.Message.Author.SendMessageAsync("", false, b.Build());
+                }
+
+                if (content.Length > 500)
+                {
+                    await Context.Message.Author.SendMessageAsync("Mod mails can not exceed 500 characters");
```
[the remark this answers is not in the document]
iShauny: fixed typos

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+                var serverConfig = _floofDb.ModMails.Find(RFURRY_SERVER_ID);
+                IGuild guild = Context.Client.GetGuild(RFURRY_SERVER_ID); // can return null
+                Discord.ITextChannel channel = await guild.GetTextChannelAsync((ulong)serverConfig.ChannelId); // can return null
+                IRole role = null;
+                if (serverConfig.ModRoleId != null)
+                {
+                    role = guild.GetRole((ulong)serverConfig.ModRoleId); // can return null
+                }
+
+                if (serverConfig == null) // not configured
```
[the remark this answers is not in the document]
iShauny: This was an issue because it was set as var instead of a specific type. Changed serverConfig to an explicit type and merged conditions

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+        public ModMailConfigModule(FloofDataContext floofDB)
+        {
+            _floofDB = floofDB;
+        }
+        private Discord.Color GenerateColor()
+        {
+            return new Discord.Color((uint)new Random().Next(0x1000000));
+        }
+
+        private void CheckServerEntryExists(ulong server)
```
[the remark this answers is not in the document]
iShauny: I've made the function retrieve the server config

Floofbot/Modules/ModMail.cs
iShauny: done

Floofbot/Modules/ModMail.cs
iShauny: I've added a bool check that is triggered when it finds a role, if it has found a role, it continues checking the role list for other roles with the same name and throws an error if there are

Floofbot/Modules/ModMail.cs
iShauny: Set a null check that will set the current channel to the channel to send modmails to

Floofbot/Modules/ModMail.cs
iShauny: fixed

Floofbot/Modules/ModMail.cs
```diff
@@ -0,0 +1,195 @@
...
+using System.Threading.Tasks;
+
+namespace Floofbot.Modules
+{
+    [Summary("Send a message directly to the server's moderators")]
+    [Group("modmail")]
+    public class ModMailModule : InteractiveBase
+    {
+        private FloofDataContext _floofDb;
+        static readonly ulong RFURRY_SERVER_ID =225980129799700481; // TODO: Replace this so that it works on more servers
```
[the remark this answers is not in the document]
iShauny: should be done

Floofbot/Modules/ModMail.cs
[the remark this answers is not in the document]
jkchen2: Ah, alright. I was thinking the cast should be unnecessary due to the check at L57, but if `GetRole` is picky about it, then it is what it is

Floofbot/Modules/ModMail.cs
```diff
@@ -118,15 +116,23 @@ private void CheckServerEntryExists(ulong server)
                     ModRoleId = null
                 });
                 _floofDB.SaveChanges();
+                return _floofDB.ModMails.Find(server);
```
jkchen2, line 119: This isn't _quite_ what I had in mind. I was thinking of using the result of `_floofDB.Add` as the function's return value. Would that not work?

Floofbot/app.config.sample
```diff
@@ -1,7 +1,7 @@
 # Floofbot Config File
 
 # The prefix to use for the bot commands
-Prefix: "."
+Prefix: ">"
```
jkchen2, line 4: Did you mean to commit this?

iShauny: nope, whoops

PR #90, opened 2020-07-11
Added the ability for users to view their own warnlog + minor refactoring

iShauny:
Users can now view their own warn logs which are DMd to them, moderators can view their own warn logs or other user's.

### DIFF
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -426,43 +426,45 @@ namespace Floofbot.Modules
         {
             SocketGuildUser selfUser = Context.Guild.GetUser(Context.Message.Author.Id); // get the guild user
             Embed embed;
-
-            if (!selfUser.GuildPermissions.BanMembers) // not a mod
+            if (string.IsNullOrEmpty(user)) // want to view their own warnlog 
             {
-                if (string.IsNullOrEmpty(user)) // want to view their own warnlog 
-                {
-                    embed = await GetWarnings(Context.Message.Author.Id, true, Context.Channel);
+                    embed = GetWarnings(Context.Message.Author.Id, true);
                     if (embed == null)
+                    {
+                        await Context.Message.Author.SendMessageAsync($"You are a good noodle. You have no warnings!");
                         return;
+                    }
                     await Context.Message.Author.SendMessageAsync("", false, embed);
                     return;
-                }
-                else
-                {
-                    await Context.Channel.SendMessageAsync("Only moderators can view the warn logs of other users.");
-                    return;
-                }
             }
             else // a mod
             {
-                if (string.IsNullOrEmpty(user)) // want to view their own warnlog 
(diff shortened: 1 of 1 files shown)


Floofbot/Modules/Administration.cs
```diff
@@ -421,105 +422,50 @@ await foreach(var messageCollection in asyncMessageCollections)
         [Alias("wl")]
         [Summary("Displays the warning log for a given user")]
         [RequireContext(ContextType.Guild)]
```
jkchen2, line 424: Kind of a bummer this prevents usage of `warnlog` in a DM with the bot. But adding logic to check if the first argument is a guild ID/name is going to be a bit messy and is more work.

Floofbot/Modules/Administration.cs
```diff
@@ -966,5 +912,142 @@ private async Task SetUserNoteForgivenStatus(UserNote un, bool status, ulong for
...
+                            await channel.SendMessageAsync($"{badUser.Username}#{badUser.Discriminator} is a good noodle. They have no warnings or user notes!");
+                            return null;
+                        }
+                    }
+                }
+                else // own users history
+                {
+                    if (formalWarnings.Count() == 0)
+                    {
+                        await channel.SendMessageAsync($"You are a good noodle. You have no warnings!");
```
jkchen2, line 963: There is some privacy leakage here. Any user who uses `warnlog` and does _not_ get this response in the channel they sent the message in will be a dead giveaway they have at least one warning. As an optional refactor, the bot should never send a response to the `warnlog` command in the same channel and instead tell the user in a DM that they have warnings or are a good noodle.

This can probably be achieved by modifying this line and moving it to the logic on [R435](https://github.com/bealsbe/Floofbot/pull/90/files#diff-2f88d76a18ee817ca5439e76a15ab026R435)

Floofbot/Modules/Administration.cs
```diff
@@ -421,105 +422,50 @@ await foreach(var messageCollection in asyncMessageCollections)
...
-            if (badUser == null) // no user, just id in database
-                builder.WithTitle($"Warnings for {user}");
-            else
-                builder.WithTitle($"Warnings for {badUser.Username}#{badUser.Discriminator}");
-            if (formalWarnings.Count() != 0) // they have warnings
+            else // a mod
             {
-                builder.AddField(":warning: | Formal Warnings:", "\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_\\_");
-                foreach (Warning warning in formalWarnings)
+                if (string.IsNullOrEmpty(user)) // want to view their own warnlog
```
jkchen2, line 448: To avoid duplicate logic, I'd suggest checking for an empty `user` string first, then in the `else` condition, check if the user is a mod trying to access another user's warnlog.

PR #91, opened 2020-07-12
Add the ability for moderators to warn/add usernotes to people not in the server

iShauny:
This allows moderators to add a warning/usernote to the database for a specific user via their user id if they are not in the server

### DIFF
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -302,7 +302,7 @@ namespace Floofbot.Modules
                 }
                 else
                 {
-                    await Context.Channel.SendMessageAsync($"?? Could not find user \"{user}\"");
+                    await Context.Channel.SendMessageAsync($"⚠️ Could not find user \"{user}\"");
                     return;
                 }
             }
@@ -378,7 +378,7 @@ namespace Floofbot.Modules
                 }
                 else
                 {
-                    await Context.Channel.SendMessageAsync($"?? Could not find user \"{user}\"");
+                    await Context.Channel.SendMessageAsync($"⚠️ Could not find user \"{user}\"");
                     return;
                 }
             }
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -293,9 +293,17 @@ namespace Floofbot.Modules
             }
 
             IUser badUser = resolveUser(user);
+            ulong uid = 0; // used if no resolved user
             if (badUser == null) {
-                await Context.Channel.SendMessageAsync($"⚠️ Could not find user \"{user}\"");
-                return;
+                if (Regex.IsMatch(user, @"\d{17,18}"))
+                {
+                    uid = Convert.ToUInt64(user);
+                }
+                else
+                {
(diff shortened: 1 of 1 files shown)


Floofbot/Modules/Administration.cs
```diff
@@ -293,9 +293,17 @@ public async Task RemoveAutoBan([Summary("user ID")] string userId)
             }
 
             IUser badUser = resolveUser(user);
+            ulong uid = 0; // used if no resolved user
             if (badUser == null) {
-                await Context.Channel.SendMessageAsync($"⚠️ Could not find user \"{user}\"");
-                return;
+                if (Regex.IsMatch(user, @"\d{17,18}"))
```
jkchen2, line 298: While it's very unlikely, it should be possible to have a user ID that's less than 17 characters long (and in the near future, more than 18). To be on the safe side, I'd run with `\d{16,}`. I see that this regex pattern is also used in the `resolveUser` helper function, so that could be changed as well.

This also applies to R369

(Side note, I think the `resolveUser` logic for a name in the format of `<name>#<discrim>` could have a bug. If a user is named `foo#bar#1234`, the logic after the passed regex check will pass in the arguments `foo` and `bar`, not `foo#bar` and `1234`. May need to resort to substrings to fix that one.)

Floofbot/Modules/Administration.cs
```diff
@@ -306,23 +314,26 @@ public async Task RemoveAutoBan([Summary("user ID")] string userId)
                 Moderator = $"{Context.User.Username}#{Context.User.Discriminator}",
                 ModeratorId = Context.User.Id,
                 Reason = reason,
-                UserId = badUser.Id,
+                UserId = (badUser != null) ? badUser.Id : uid,
```
jkchen2, line 317: Optional refactor: Cool trick when searching up the question marks in C# code, you can use the [null coalescing operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-coalescing-operator) to condense this to `UserId = badUser?.Id ?? uid` (though I always find it makes the code look very confused haha)

jkchen2: Reading up on the rest of the changes, it seems like it might be beneficial to pull this logic out and give it its own variable as to prevent logic duplication. You could probably add an `else` condition to R307 to assign the `badUser.Id` value to `uid`

This should apply for the duplicates at R336, R387, and R394 as well.

Floofbot/Modules/Administration.cs
```diff
@@ -294,17 +294,22 @@ public async Task RemoveAutoBan([Summary("user ID")] string userId)
...
+            if (badUser == null)
+            {
                 if (Regex.IsMatch(user, @"\d{16,}"))
                 {
                     uid = Convert.ToUInt64(user);
                 }
                 else
                 {
-                    await Context.Channel.SendMessageAsync($"⚠️ Could not find user \"{user}\"");
+                    await Context.Channel.SendMessageAsync($"?? Could not find user \"{user}\"");
```
jkchen2, line 305: Minor, but why not use the ⚠️ emoji?

PR #93, opened 2020-08-21
Improved Error Logging Readability and Increase Cache Size

iShauny:
The reporting of errors has been improved in readability and the message cache size has been increased to 1000.


Floofbot/Handlers/CommandHandler.cs
```diff
@@ -104,11 +105,14 @@ private async Task HandleCommandAsync(SocketMessage s)
 
             if (msg.HasStringPrefix(prefix, ref argPos) || msg.HasMentionPrefix(_client.CurrentUser, ref argPos))
             {
+                if (msg.Author.IsBot)
```
jkchen2, line 108: Why not move this check up to R89?

PR #94, opened 2020-08-22
Added Basic Raid Protection Service and Commands

iShauny:
The rules for triggering the raid protection are:
- Over 10 Mass mention spams (instant ban)
- Over 3 user strikes within 5 mins
- - Can gain from spamming 5 messages in 5 seconds
- - Can gain from posting invite links
- - Can gain from spamming the same letter over 5 times in a row
- - Can gain from spamming 5 emojis both custom and normal
- After 5 minutes, the user is forgiven

Mod alerts when
- Over 3 users have been punished within 30 mins
- Over 5 users joined in 2 mins
- A user has been autobanned for mention spam

Options:
- Toggle on/off the raid protection
- Set a muted role
- Set a mod role to ping for alerts
- Set a mod channel to send mod alerts to
- Toggle on/off banning instead of muting (mute role must be set to disable this)
- Add an exceptions role which is not affected by raid protection

### DIFF
--- a/Floofbot/Modules/RaidProtectionCommands.cs
+++ b/Floofbot/Modules/RaidProtectionCommands.cs
@@ -56,16 +56,11 @@ namespace Floofbot.Modules
         private async Task<SocketRole> resolveRole(string input, SocketGuild guild, ISocketMessageChannel channel)
         {
             SocketRole role = null;
-            bool roleFound = false;
             //resolve roleID or @mention
             if (Regex.IsMatch(input, @"\d{10,}"))
             {
                 string roleID = Regex.Match(input, @"\d{10,}").Value;
                 role = guild.GetRole(Convert.ToUInt64(roleID));
-                if (role != null)
-                    roleFound = true;
-                else
-                    roleFound = false;
             }
             //resolve role name
             else
@@ -74,9 +69,8 @@ namespace Floofbot.Modules
                 {
                     if (r.Name.ToLower() == input.ToLower())
                     {
-                        if (roleFound == false) // ok we found 1 role thats GOOD
+                        if (role == null) // ok we found 1 role thats GOOD
                         {
-                            roleFound = true;
                             role = r;
                         }
                         else // there is more than 1 role with the same name!
--- a/Floofbot/Modules/RaidProtectionCommands.cs
+++ b/Floofbot/Modules/RaidProtectionCommands.cs
@@ -87,14 +87,7 @@ namespace Floofbot.Modules
                     }
                 }
             }
-            if (roleFound)
-            {
-                return role;
-            }
-            else
-            {
-                return null;
-            }
+             return role;
         }
         [Command("modchannel")]
         [Summary("OPTIONAL: Sets the mod channel for raid notifications")]
--- a/Floofbot/Modules/RaidProtectionCommands.cs
+++ b/Floofbot/Modules/RaidProtectionCommands.cs
@@ -100,7 +100,7 @@ namespace Floofbot.Modules
                 var ServerConfig = GetServerConfig(Context.Guild.Id);
                 ServerConfig.BanOffenders = !ServerConfig.BanOffenders;
                 _floofDB.SaveChanges();
-                await Context.Channel.SendMessageAsync("Raid Protection Bans " + (ServerConfig.Enabled ? "Enabled!" : "Disabled!"));
+                await Context.Channel.SendMessageAsync("Raid Protection Bans " + (ServerConfig.BanOffenders ? "Enabled!" : "Disabled!"));
             }
             catch (Exception ex)
             {
@@ -193,7 +193,7 @@ namespace Floofbot.Modules
 
                 if (mutedRoleFound)
                 {
-                    ServerConfig.ModRoleId = roleId;
+                    ServerConfig.MutedRoleId = roleId;
                     _floofDB.SaveChanges();
                     await Context.Channel.SendMessageAsync("Muted role set!");
                 }
@@ -243,7 +243,7 @@ namespace Floofbot.Modules
 
                 if (exceptionsRoleFound)
                 {
-                    ServerConfig.ModRoleId = roleId;
+                    ServerConfig.ExceptionRoleId = roleId;
                     _floofDB.SaveChanges();
                     await Context.Channel.SendMessageAsync("Exceptions role set!");
                 }
--- a/Floofbot/Modules/RaidProtectionCommands.cs
+++ b/Floofbot/Modules/RaidProtectionCommands.cs
@@ -165,17 +165,17 @@ namespace Floofbot.Modules
             var ServerConfig = GetServerConfig(Context.Guild.Id);
             var foundRole = await resolveRole(role, Context.Guild, Context.Channel);
 
-                if (foundRole != null)
-                {
-                    ServerConfig.ModRoleId = foundRole.Id;
-                    _floofDB.SaveChanges();
-                    await Context.Channel.SendMessageAsync("Mod role set to " + foundRole.Name + "!");
-                }
-                else
-                {
-                    await Context.Channel.SendMessageAsync("Unable to set that role.");
-                    return;
-                }
+            if (foundRole != null)
+            {
+                ServerConfig.ModRoleId = foundRole.Id;
+                _floofDB.SaveChanges();
+                await Context.Channel.SendMessageAsync("Mod role set to " + foundRole.Name + "!");
+            }
+            else
+            {
+                await Context.Channel.SendMessageAsync("Unable to set that role.");
+                return;
+            }
         }
         [Command("mutedrole")]
         [Summary("OPTIONAL: A role to give to users to mute them in the server. If not set, users are banned by default.")]
--- a/Floofbot/Modules/RaidProtectionCommands.cs
+++ b/Floofbot/Modules/RaidProtectionCommands.cs
@@ -1,0 +1,263 @@
+﻿using Discord;
+using Discord.Addons.Interactive;
+using Discord.Commands;
+using Discord.WebSocket;
+using Floofbot.Services.Repository;
+using Floofbot.Services.Repository.Models;
+using Serilog;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Floofbot.Modules
+{
+    [Summary("Raid protection configuration commands")]
+    [RequireUserPermission(GuildPermission.Administrator)]
+    [Group("raidconfig")]
+    public class RaidProtectionCommands : InteractiveBase
+    {
+        private FloofDataContext _floofDB;
+
+        public RaidProtectionCommands(FloofDataContext floofDB)
+        {
+            _floofDB = floofDB;
+        }
+        private Discord.Color GenerateColor()
+        {
+            return new Discord.Color((uint)new Random().Next(0x1000000));
+        }
+
+        private RaidProtectionConfig GetServerConfig(ulong server)
+        {
+            // checks if server exists in database and adds if not
+            var serverConfig = _floofDB.RaidProtectionConfigs.Find(server);
+            if (serverConfig == null)
+            {
+                _floofDB.Add(new RaidProtectionConfig
+                {
+                    ServerId = server,
+                    Enabled = false,
+                    ModChannelId = null,
+                    ModRoleId = null,
+                    MutedRoleId = null,
+                    BanOffenders = true,
+                    ExceptionRoleId = null
+                });
+                _floofDB.SaveChanges();
+                return _floofDB.RaidProtectionConfigs.Find(server);
+            }
+            else
+            {
+                return serverConfig;
+            }
+        }
+
+        [Command("modchannel")]
+        [Summary("OPTIONAL: Sets the mod channel for raid notifications")]
+        public async Task ModChannel([Summary("Channel (eg #alerts)")]Discord.IChannel channel = null)
+        {
+            if (channel == null)
+            {
+                channel = Context.Channel;
+            }
+            var ServerConfig = GetServerConfig(Context.Guild.Id);
+            ServerConfig.ModChannelId = channel.Id;
+            _floofDB.SaveChanges();
+            await Context.Channel.SendMessageAsync("Channel updated! I will raid notifications to <#" + channel.Id + ">");
+        }
+
+        [Command("toggle")]
+        [Summary("Toggles the raid protection")]
+        public async Task Toggle()
+        {
+
+            // try toggling
+            try
+            {
+                // check the status of raid protection
+                var ServerConfig = GetServerConfig(Context.Guild.Id);
+                ServerConfig.Enabled = !ServerConfig.Enabled;
+                _floofDB.SaveChanges();
+                await Context.Channel.SendMessageAsync("Raid Protection " + (ServerConfig.Enabled ? "Enabled!" : "Disabled!"));
+            }
+            catch (Exception ex)
+            {
+                await Context.Channel.SendMessageAsync("An error occured: " + ex.Message);
+                Log.Error("Error when trying to toggle raid protection: " + ex);
+                return;
+            }
+        }
+        [Command("togglebans")]
+        [Summary("Toggles whether to ban or mute users that trigger the raid detection.")]
+        public async Task ToggleBans()
+        {
+
+            // try toggling
+            try
+            {
+                // check the status of raid protection
+                var ServerConfig = GetServerConfig(Context.Guild.Id);
+                ServerConfig.BanOffenders = !ServerConfig.BanOffenders;
+                _floofDB.SaveChanges();
+                await Context.Channel.SendMessageAsync("Raid Protection Bans " + (ServerConfig.Enabled ? "Enabled!" : "Disabled!"));
+            }
+            catch (Exception ex)
+            {
+                await Context.Channel.SendMessageAsync("An error occured: " + ex.Message);
+                Log.Error("Error when trying to toggle raid protection banning: " + ex);
+                return;
+            }
+        }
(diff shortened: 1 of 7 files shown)


Floofbot/Modules/RaidProtectionCommands.cs
```diff
@@ -0,0 +1,231 @@
...
+            }
+            else
+            {
+                return serverConfig;
+            }
+        }
+        private async Task<SocketRole> resolveRole(string input, SocketGuild guild, ISocketMessageChannel channel)
+        {
+            SocketRole role = null;
+            bool roleFound = false;
```
CryptidVulpes, line 59: This boolean is a bit redundant. You can just check if `role == null` because if role is set, it has been found.

Floofbot/Modules/RaidProtectionCommands.cs
```diff
@@ -0,0 +1,231 @@
...
+                        }
+                        else // there is more than 1 role with the same name!
+                        {
+                            await channel.SendMessageAsync("More than one role exists with that name! Not sure what to do! Please resolve this. Aborting..");
+                            return null;
+                        }
+                    }
+                }
+            }
+            if (roleFound)
```
CryptidVulpes, line 90: Here you can always `return role` because can either be a legit role or `null`.

Floofbot/Modules/RaidProtectionCommands.cs
```diff
@@ -0,0 +1,231 @@
...
+                });
+                _floofDB.SaveChanges();
+                return _floofDB.RaidProtectionConfigs.Find(server);
+            }
+            else
+            {
+                return serverConfig;
+            }
+        }
+        private async Task<SocketRole> resolveRole(string input, SocketGuild guild, ISocketMessageChannel channel)
```
CryptidVulpes, line 56: Should the type be `Task<SocketRole?>` because we are also returning `null`?

Floofbot/Modules/RaidProtectionCommands.cs
```diff
@@ -0,0 +1,231 @@
...
+            if (role == null)
+            {
+                await Context.Channel.SendMessageAsync("", false, new EmbedBuilder { Description = $"💾 Usage: `raidconfig modrole [rolename]`", Color = GenerateColor() }.Build());
+                return;
+            }
+
+            var ServerConfig = GetServerConfig(Context.Guild.Id);
+            var foundRole = await resolveRole(role, Context.Guild, Context.Channel);
+
+                if (foundRole != null)
```
CryptidVulpes, line 168: Indentation looks a bit broken here

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+        }
+        public async Task NotifyModerators(SocketRole modRole, ITextChannel modChannel, string message)
+        {
+            await modChannel.SendMessageAsync(modRole.Mention + " there may be a possible raid! Reason: ``" + message + "``");
+        }
+        private async void UserPunishmentTimeout(ulong userid)
+        {
+            await Task.Delay(forgivenDuration);
+            if (userPunishmentCount.ContainsKey(userid))
+                if (userPunishmentCount[userid] != 0)
```
CryptidVulpes, line 67: These `if`s can be evaluated on the same line

CryptidVulpes: Same thing occurs twice below

CryptidVulpes (commented):
This only a partial review again.

Floofbot/Modules/RaidProtectionCommands.cs
iShauny: Fixed

Floofbot/Services/RaidProtectionService.cs
iShauny: If I ran these on the same line the second if statement will return null if they key does not exist

CryptidVulpes: I'm not sure if I follow what you are trying to say, but
`if (userPunishmentCount.ContainsKey(userid) && userPunishmentCount[userid] != 0)` should work, because if `userPunishmentCount.ContainsKey(userid)` evaluates to `false` (I'm assuming `null == false`), the second statement is not evaluated.

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+            {
+                userMessageCount.Add(msg.Author.Id, 1);
+            }
+            // we run an async task to remove their point after the specified duration
+            UserMessageCountTimeout(msg.Author.Id);
+            return false;
+        }
+        public async Task<bool> CheckMentions(SocketUserMessage msg, IGuild guild, SocketRole modRole, ITextChannel modChannel)
+        {
+            if (msg.MentionedUsers.Count > 10)
```
jkchen2, line 131: This should probably be configurable

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+using System.Text.RegularExpressions;
+using System.Threading.Channels;
+using System.Threading.Tasks;
+
+namespace Floofbot.Services
+{
+    public class RaidProtectionService
+    {
+        // will store user id and how many counts they've had
+        private Dictionary<ulong, int> userPunishmentCount = new Dictionary<ulong, int>();
```
jkchen2, line 24: `userPunishmentCount`, `userMessageCount` and `punishedUsers` are all not tied to any one guild. Thus, a user can be preemptively "primed" for a strike on a server they weren't active on. Behavior in `CheckMessage` will need to be changed as a result.

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+        public async Task NotifyModerators(SocketRole modRole, ITextChannel modChannel, string message)
+        {
+            await modChannel.SendMessageAsync(modRole.Mention + " there may be a possible raid! Reason: ``" + message + "``");
+        }
+        private async void UserPunishmentTimeout(ulong userid)
+        {
+            await Task.Delay(forgivenDuration);
+            if (userPunishmentCount.ContainsKey(userid))
+                if (userPunishmentCount[userid] != 0)
+                    userPunishmentCount[userid] -= 1;
```
jkchen2, line 68: Optional goal: to keep things running lean, consider removing keys from these dictionaries (and the other dictionaries)

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+                        userPunishmentCount[msg.Author.Id] += 1;
+                    }
+                    else // they were a good boye but now they are not
+                    {
+                        userPunishmentCount.Add(msg.Author.Id, 1);
+                    }
+                    // reset the count after punishing the user
+                    userMessageCount[msg.Author.Id] = 0;
+                    // we run an async task to remove their point after the specified duration
+                    await Task.Run(async () =>
```
jkchen2, line 109: Optional goal: instead of having one timer for each strike point, have one timer for all strikes. That is, the timer is reset if another strike occurs before the timer finishes. This prevents users from floating at a certain strike level indefinitely (they receive strikes at the same rate they lose them).

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+                    // we return here because we only need to check for at least one match, doesnt matter if there are more
+                    return true;
+                }                     
+            }
+            return false;
+
+        }
+        public async Task<bool> CheckInviteLinks(SocketMessage msg)
+        {
+            if (msg.Content.Contains("discord.gg") || msg.Content.Contains("d.gg"))
```
jkchen2, line 186: This is a little too wide of a net (I'd also like to point out that the first half of this statement is redundant)

I'll have to come back and come up with a better way to parse out invite links, but be aware that this would catch an innocent https://discord.gg link and miss https://discord.com/invite/rfurry

Floofbot/Modules/RaidProtectionCommands.cs
iShauny: This is not the purpose of this boolean. It is used to check if multiple roles of the same name are found. Using role will overwrite the found role and will cause issues with multiple roles. By setting a flag when a role is found, i can check the state of that flag if another role is found. And if I just take the first instance of role, it will ignore duplicate roles and may not grab the right one

Floofbot/Modules/RaidProtectionCommands.cs
iShauny: done

Floofbot/Modules/RaidProtectionCommands.cs
```diff
@@ -0,0 +1,231 @@
...
+            // try toggling
+            try
+            {
+                // check the status of raid protection
+                var ServerConfig = GetServerConfig(Context.Guild.Id);
+                ServerConfig.Enabled = !ServerConfig.Enabled;
+                _floofDB.SaveChanges();
+                await Context.Channel.SendMessageAsync("Raid Protection " + (ServerConfig.Enabled ? "Enabled!" : "Disabled!"));
+            }
+            catch (Exception ex)
```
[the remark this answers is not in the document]
iShauny: This was actually legacy code from a copy/paste and can be removed due to the global error handling

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,387 @@
...
+{
+    public class RaidProtectionService
+    {
+        // will store user id and how many counts they've had
+        private Dictionary<ulong, int> userPunishmentCount = new Dictionary<ulong, int>();
+        // used to keep track of the number of messages a user sent for spam protection
+        private Dictionary<ulong, int> userMessageCount = new Dictionary<ulong, int>();
+        // a list of punished users used to detect any potential raids
+        private List<SocketUser> punishedUsers = new List<SocketUser>();  
+        // determines how long before a user is forgiven for a punishment
```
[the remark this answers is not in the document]
iShauny: I considered that, but I do not want to clog up the config file with parameters that do not usually need to change. Personally, I think I prefer having standard values that the user can modify if they choose in the code.

Floofbot/Services/RaidProtectionService.cs
iShauny: done

Floofbot/Services/RaidProtectionService.cs
iShauny: Changed to make them guild dependent

Floofbot/Services/RaidProtectionService.cs
iShauny: Done

Floofbot/Services/RaidProtectionService.cs
iShauny: Added a regex algorithm for this

Floofbot/Modules/RaidProtectionCommands.cs
iShauny: Changed

Floofbot/Modules/RaidProtectionCommands.cs
```diff
@@ -0,0 +1,231 @@
...
+                await Context.Channel.SendMessageAsync("An error occured: " + ex.Message);
+                Log.Error("Error when trying to toggle raid protection banning: " + ex);
+                return;
+            }
+        }
+        [Command("modrole")]
+        [Summary("OPTIONAL: A role to ping for raid alerts.")]
+        public async Task SetModRole(string role = null)
+        {
+            if (role == null)
```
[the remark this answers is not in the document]
iShauny: I have changed it so that if you do not provide a role, the role is cleared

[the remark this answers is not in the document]
iShauny: This also means that for mods, pings will stop
for exceptions, users wont be exempt
for mutes, it will default to banning

Floofbot/Services/RaidProtectionService.cs
iShauny: This is substantially harder, and I think I would prefer this left as a future PR

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,419 @@
...
+            if (serverConfig.ExceptionRoleId != null)
+            {
+                // returns null if exception role doe not exist anymore
+                var exceptionsRole = guild.GetRole((ulong)serverConfig.ExceptionRoleId); 
+                var guildUser = guild.GetUser(userMsg.Author.Id);
+                // role must exist and user must exist in server
+                if (exceptionsRole != null && guildUser != null)
+                {
+                    foreach (IRole role in guildUser.Roles)
+                        {
```
jkchen2, line 331: Style: Some wonky indentation here

jkchen2 (changes requested):
Just as a general comment, it'd be nice to have more log statements to track things that would otherwise provide no feedback. It'd be useful to debug strange edge cases when we eventually roll out raid protection.

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,419 @@
...
+                    // we return here because we only need to check for at least one match, doesnt matter if there are more
+                    return true;
+                }                     
+            }
+            return false;
+
+        }
+        private bool CheckInviteLinks(SocketMessage msg, ulong guildId)
+        {
+            var regex = "(https?:\\/\\/)?(www\\.)?((d(iscord)?\\.gg)|(discord(app)?\\.com))/+\\w{5,}/?";
```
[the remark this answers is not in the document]
iShauny: Updated

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,419 @@
...
+                return;
+            // increment user join count for guild
+            if (numberOfJoins.ContainsKey(guild))
+            {
+                numberOfJoins[guild] += 1;
+                if (numberOfJoins[guild] >= maxNumberOfJoins)
+                {
+                    var modRole = (serverConfig.ModRoleId != null) ? server.GetRole((ulong)serverConfig.ModRoleId) : null;
+                    var modChannel = (serverConfig.ModChannelId != null) ? server.GetChannel((ulong)serverConfig.ModChannelId) as ITextChannel : null;
+                    await NotifyModerators(modRole, modChannel, "Excessive number of new joins in short time period.");
```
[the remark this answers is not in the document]
iShauny: Added a null check inside the NotifyModerators function

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,419 @@
...
+                    //sends message to user
+                    EmbedBuilder builder = new EmbedBuilder();
+                    builder.Title = "⚖️ Ban Notification";
+                    builder.Description = $"You have been banned from {guild.Name}";
+                    builder.AddField("Reason", reason);
+                    builder.Color = Discord.Color.Red;
+                    await msg.Author.SendMessageAsync("", false, builder.Build());
+                    await guild.AddBanAsync(msg.Author, 0, reason);
+
+                    await NotifyModerators(modRole, modChannel, " I have banned a user for mentioning " + msg.MentionedUsers.Count() + " members");
```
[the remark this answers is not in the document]
iShauny: Added a null check inside the NotifyModerators function

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,419 @@
...
+                UserPunishmentTimeout(guildId, msg.Author.Id);
+
+                SendMessageAndDelete(msg.Author.Mention + " do not spam emojis!", msg.Channel);
+
+                return true;
+            }
+            else
+                return false;
+        }
+        public async Task CheckForExcessiveJoins(IGuild guild)
```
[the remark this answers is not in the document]
iShauny: Merged a bunch of if statements and fixed copy/paste error

Floofbot/Services/RaidProtectionService.cs
iShauny: Fixed

jkchen2 (approved):
I think it's okay to be merged, but I still urge you to add more debug statements.

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,427 @@
...
+        // These are used to determine if there are an excessive number of joins in a short time frame
+        private static int maxNumberOfJoins = 5;
+        private static int userJoinsDelay = 2 * 60 * 1000; // 2 min
+        private Dictionary<IGuild, int> numberOfJoins = new Dictionary<IGuild, int>();
+        // The max number of repeated emojis before triggering the raid protection
+        private static int maxNumberEmojis = 5;
+
+
+
+        public RaidProtectionService()
```
jkchen2, line 56: Can the implicit constructor be used instead?

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -0,0 +1,427 @@
...
+        private static int removePunishedUserDelay = 30 * 60 * 1000; // 30 min
+        // the number of punished users within a time frame before the mods are alerted of possible raids
+        private static int maxNumberPunishedUsers = 3;
+        // These are used to determine if there are an excessive number of joins in a short time frame
+        private static int maxNumberOfJoins = 5;
+        private static int userJoinsDelay = 2 * 60 * 1000; // 2 min
+        private Dictionary<IGuild, int> numberOfJoins = new Dictionary<IGuild, int>();
+        // The max number of repeated emojis before triggering the raid protection
+        private static int maxNumberEmojis = 5;
+
```
jkchen2, line 53: Random extra whitespace lines here and on L25/L26

PR #96, opened 2020-08-22
Added Support To Specify App Config in Command Line Argument

iShauny:
Also added the database file path to the config file


Floofbot/app.config.sample
```diff
@@ -21,6 +21,9 @@ Token: ""
 # Set this to a valid Server ID
 ModMailServer: ""
 
+# Path to the SQLite database file
+DbPath: ""
```
jkchen2, line 25: May be a good idea to set a default path (`floofData.db`)

PR #98, opened 2020-08-24
Add basic build action

jkchen2:
This builds binaries for 3 targets (all x86-64):

* Windows
* Linux
* Linux (musl libc)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,45 @@
+name: Build binaries
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


iShauny (approved):

CryptidVulpes (commented):
Should we build a debug binary too, just so that it's easier to debug a running version of the bot, if it ever comes to that?

jkchen2:
> Should we build a debug binary too, just so that it's easier to debug a running version of the bot, if it ever comes to that?

Good question. My intention with producing build artifacts is to be able to upload them directly to a GitHub release. While it'd be pretty easy to add another set of 3 debug targets, it'd significantly increase the storage space used and build time. While neither of those resources seem to be at risk of hitting any quota for a small project like this, I'd rather keep it small if possible.

Personally, I think it'd be more reasonable for us to make a debug build manually if we ever needed to debug in production (which we shouldn't be doing anyway), but in the end it could go either way for me. If you're confident it'd be beneficial, I can add it in.

PR #99, opened 2020-08-24
Created Basic Backup Routine

iShauny:
A c# backup routine has been provided
Two basic SQLite backup scripts have also been provided
Config file has been updated to reflect these changes

### DIFF
--- a/Floofbot/Services/BackupService.cs
+++ b/Floofbot/Services/BackupService.cs
@@ -30,7 +30,6 @@ namespace Floofbot.Services
             while (true)
             {
                 double targetDelay = backupTime.TotalSeconds - DateTime.UtcNow.TimeOfDay.TotalSeconds;
-                Console.WriteLine(targetDelay.ToString());
                 if (targetDelay < 0)
                 {
                     targetDelay += 86400;
--- a/Floofbot/Services/BackupService.cs
+++ b/Floofbot/Services/BackupService.cs
@@ -1,8 +1,9 @@
 ﻿using Floofbot.Configs;
 using Serilog;
 using System;
-using System.Collections.Generic;
-using System.Text;
+using System.IO;
+using System.Runtime.InteropServices;
+using System.Threading;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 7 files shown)


Floofbot/Services/BackupService.cs
```diff
@@ -0,0 +1,82 @@
...
+                        catch (FileNotFoundException)
+                        {
+                            Log.Error("The backup script file could not be found. Backup not successful.");
+                            return;
+                        }
+                        catch (Exception e)
+                        {
+                            Log.Fatal("Exception occured when trying to backup the the database: " + e);
+                        }
+                        Thread.Sleep(60 * 1000); // wait 1 min to prevent backing up multiple times
```
jkchen2, line 76: This is not outside the `if` statement, which means the loop is busy waiting.

jkchen2 (changes requested):
I highly recommend that you take a different approach to backups. Currently, the bot itself dictates the backup numbers. This means that if the last backup made by the bot was `backupIndex == 1`, and the bot crashes, the next backup will overwrite the latest one.

As an extension to that, the scripts should simply increment backup files by one for each execution.

For example, if we run the backup script 4 times with a maximum of 3 backups configured, this is what I'd expect:

Iteration 1:
```
backup1.db <- This is the new backup file
```

Iteration 2:
```
backup1.db <- This is the new backup file
backup2.db <- This is the previous backup file, now renamed
```

Iteration 3:
```
backup1.db <- This is the new backup file
backup2.db
backup3.db <- This is the same backup1.db file from the first iteration, and backup2.db from the second iteration
```

Iteration 4:
```
backup1.db <- This is the new backup file
backup2.db
backup3.db <- This is backup2.db from iteration 3. The previous backup3.db is now overwritten
```

One last thing, this is a fairly inefficient way to schedule a task. You can calculate when the next backup is scheduled to happen, then initiate a `Task.Delay` (`Thread.Sleep` blocks the running thread) for the remaining time.

jkchen2 (approved):

PR #109, opened 2020-09-23
Refine prefix checking and problematic command name stripping

jkchen2:


jkchen2:
Rebased because I accidentally left an extra space in b115e9f

jkchen2:
Another force push because I realized there was a much easier way to get the command name by itself.

jkchen2:
Another rebase to remove an unused import that slipped through 😅

iShauny (approved):

PR #119, opened 2020-10-01
Added regex into a group to allow phrases with spaces

iShauny:
This allows for filtered phrases rather than just words


jkchen2 (changes requested):
This doesn't work. I think you meant to keep the dollar sign.

jkchen2 (approved):

PR #122, opened 2020-10-02
XKCD now gets latest on no argument

the-og-gear:
Previously, the XKCD command would just spit out an error message with a
possibly unknown value to some. I changed this around so that it now
gets the latest XKCD comic with no argument given (as it says in the
help command and module summary), as well as gives the actual integer
value for the maximum comic ID.

### DIFF
--- a/Floofbot/Modules/Fun.cs
+++ b/Floofbot/Modules/Fun.cs
@@ -32,19 +32,27 @@ namespace Floofbot.Modules
         public async Task XKCD([Summary("Comic ID")] string comicId = "")
         {
             int parsedComicId;
-            if (!int.TryParse(comicId, out parsedComicId) || parsedComicId <= 0)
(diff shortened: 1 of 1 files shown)


Floofbot/Modules/Fun.cs
```diff
@@ -32,19 +32,27 @@ public async Task AskEightBall([Summary("question")][Remainder] string question)
         public async Task XKCD([Summary("Comic ID")] string comicId = "")
         {
             int parsedComicId;
-            if (!int.TryParse(comicId, out parsedComicId) || parsedComicId <= 0)
+            if ((!int.TryParse(comicId, out parsedComicId) || parsedComicId <= 0) && !comicId.ToLower().Equals(""))
```
jkchen2, line 35: For readability I'd suggest using `String.IsNullOrEmpty` instead

the-og-gear: Ah, good point, will push that change once I get back to the laptop.

jkchen2 (approved):

PR #127, opened 2020-11-25
Updated Raid Protection To Include Timestamps

iShauny:
- Several parameters have been set to configurable variables
- Users last sent message in a server is now recorded to compare timestamps
- This means that if there is API lag or the discord bot lags, it will not trigger false positives in raid protection
- Word filter service is implemented into raid protection

### DIFF
--- a/Floofbot/Services/RaidProtectionService.cs
+++ b/Floofbot/Services/RaidProtectionService.cs
@@ -3,20 +3,11 @@ using Discord.WebSocket;
 using Floofbot.Configs;
 using Floofbot.Services.Repository;
 using Floofbot.Services.Repository.Models;
-using Microsoft.AspNetCore.Http;
-using Microsoft.AspNetCore.SignalR;
-using Microsoft.AspNetCore.WebUtilities;
 using Serilog;
 using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Diagnostics.Eventing.Reader;
 using System.Linq;
-using System.Net.NetworkInformation;
-using System.Security.Policy;
-using System.Text;
 using System.Text.RegularExpressions;
-using System.Threading.Channels;
 using System.Threading.Tasks;
 
 namespace Floofbot.Services
--- a/Floofbot/Services/RaidProtectionService.cs
+++ b/Floofbot/Services/RaidProtectionService.cs
@@ -29,6 +29,8 @@ namespace Floofbot.Services
         private Dictionary<ulong, Dictionary<ulong, int>> userPunishmentCount = new Dictionary<ulong, Dictionary<ulong, int>>();
         // used to keep track of the number of messages a user sent for spam protection
         private Dictionary<ulong, Dictionary<ulong, int>> userMessageCount = new Dictionary<ulong, Dictionary<ulong, int>>();
+        // used to track the last message a user has sent in the server
+        private Dictionary<ulong, Dictionary<ulong, SocketMessage>> lastUserMessageInGuild = new Dictionary<ulong, Dictionary<ulong, SocketMessage>>();
         // a list of punished users used to detect any potential raids
         private Dictionary<ulong, List<SocketUser>> punishedUsers = new Dictionary<ulong, List<SocketUser>>();
         // contains the number of joins in a guild in a short time frame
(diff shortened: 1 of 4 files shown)


Floofbot/Services/RaidProtectionService.cs
```diff
@@ -61,6 +60,15 @@ public RaidProtectionService()
             userJoinsDelay = raidConfig["UserJoinsDelay"];
             maxNumberEmojis = raidConfig["MaxNumberEmojis"];
             maxNumberSequentialCharacters = raidConfig["MaxNumberSequentialCharacters"];
+            durationBetweenMessages = raidConfig["DurationBetweenMessages"];
+            maxMessageSpam = raidConfig["MaxMessageSpam"];
+        }
+        private async Task HandleBadMessage(SocketUser user, SocketMessage msg)
```
jkchen2, line 66: Looks like this isn't being called anywhere (I'm assuming this is supposed to be used in place of R161?)

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -130,13 +137,63 @@ private void ensureGuildInDictionaries(ulong guildId)
...
+                    lastUserMessageInGuild[guildId][msg.Author.Id] = msg;
+                }
+                // record last user message in server
+                else
+                {
+                    lastUserMessageInGuild[guildId].Add(msg.Author.Id, msg);
+                }
+
+                // compare timestamps of messages
+                TimeSpan timeBetweenMessages = msg.Timestamp - lastUserMessageInGuild[guildId][msg.Author.Id].Timestamp;
```
jkchen2, line 190: If the `lastUserMessageInGuild` dictionary doesn't contain a user's last message (i.e. this is the first message the bot receives), wouldn't this always be 0 and trigger the following `if` statement?

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -130,13 +137,63 @@ private void ensureGuildInDictionaries(ulong guildId)
...
+                if (timeBetweenMessages.TotalSeconds < durationBetweenMessages)
+                    userMessageCount[guildId][msg.Author.Id] += 1;
+                else
+                    return false;
+
+                if (userMessageCount[guildId][msg.Author.Id] >= maxMessageSpam) // no more than 2 messages in time frame
```
jkchen2, line 196: Outdated comment

Floofbot/Services/RaidProtectionService.cs
```diff
@@ -372,7 +431,7 @@ public async Task<bool> CheckMessage(FloofDataContext _floofDb, SocketMessage ms
 
             if (spammedMentions)
                 return false; // user already banned
-            if (userSpammedMessages || userSpammedLetters || userSpammedInviteLink || userSpammedEmojis)
+            if (filteredWord || userSpammedMessages || userSpammedLetters || userSpammedInviteLink || userSpammedEmojis)
```
jkchen2, line 434: This got me thinking, what if warnings compounded on each other? For example, if a user spammed a filtered word _and_ messages, should that count double?

Also there could be a concept of infraction weights. Spamming emojis is a less severe action than triggering the word filter.

jkchen2 (changes requested):

Floofbot/Services/RaidProtectionService.cs
iShauny: Removed

Floofbot/Services/RaidProtectionService.cs
iShauny: Updated

Floofbot/Services/RaidProtectionService.cs
iShauny: Added in a return statement after the message is added into the dictionary. If this is the first message, we might as well return because they clearly aren't spamming if its the first one

Floofbot/Services/RaidProtectionService.cs
iShauny: I think these can be added as an issue and fixed in a future PR

jkchen2 (approved):

PR #129, opened 2020-11-26
Axel/conversions

Shelme74:
This new module hosts code for the `.convert` command. More info about its expected behavior and usage will be explained in the Discord server.


Floofbot/Modules/Conversions.cs
```diff
@@ -0,0 +1,165 @@
...
+            string celPat = "\\b(\\d+)(c)\\b";
+            Regex celReg = new Regex(celPat, RegexOptions.IgnoreCase);
+            string miPat = "\\b(\\d+)(mi)\\b";
+            Regex miReg = new Regex(miPat, RegexOptions.IgnoreCase);
+            string kmPat = "\\b(\\d+)(km)\\b";
+            Regex kmReg = new Regex(kmPat, RegexOptions.IgnoreCase);
+            string kgPat = "\\b(\\d+)(kg)\\b";
+            Regex kgReg = new Regex(kgPat, RegexOptions.IgnoreCase);
+            string lbPat = "\\b(\\d+)(lb)||(lbs)\\b";
+            Regex lbReg = new Regex(lbPat, RegexOptions.IgnoreCase);
```
iShauny, line 34: Might be worth just compressing these lines, not sure it needs to be declared as a string first, could also just add these into an array of some sort and iterate it. Seems excessive to have so many if statements and repeating code. Either way, I think it would be best if you could compress this code a little

Floofbot/Modules/Conversions.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            if (fahReg.Match(input).Success)
```
iShauny, line 36: Might be worth adding an else statement. One that lists the available conversions.

Floofbot/Modules/Conversions.cs
```diff
@@ -0,0 +1,165 @@
+﻿using System;
+using System.Threading.Tasks;
+using System.Text.RegularExpressions;
+using Discord;
+using Discord.Addons.Interactive;
+using Discord.Commands;
+using UnitsNet;
+
+namespace Floofbot
```
iShauny, line 9: Does this need to be an entire module? I think it would be better in utilities

iShauny (changes requested):

Floofbot/Modules/Conversions.cs
Shelme74: I just fixed that, along with compressing the code.

Floofbot/Modules/Conversions.cs
Shelme74: Might be worth moving it to utilities, I coded it in a different file for clarity on my end.

Shelme74:
Many commits have appeared here, I just rebased the branch and didn't think it would have shown up here

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
             }, true);
         }
     }
+    [Summary("Conversion commands")]
+    [Discord.Commands.Name("Conversions")]
+    public class Conversions : InteractiveBase
+    {
+        private static readonly Discord.Color EMBED_COLOR = Color.Magenta;
```
iShauny, line 265: Not needed, there's already a readonly static at the top

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
             }, true);
         }
     }
+    [Summary("Conversion commands")]
+    [Discord.Commands.Name("Conversions")]
+    public class Conversions : InteractiveBase
```
iShauny, line 263: I'm not sure this needs to be a class, can just be a command

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
...
+            Regex lbReg = new Regex("\\b(\\d+)(lbs)\\b", RegexOptions.IgnoreCase);
+
+            if (fahReg.Match(input).Success)
+            {
+                Match m = fahReg.Match(input);
+
+                Group g = m.Groups[1];
+
+                string FahStr = Convert.ToString(g);
+                double FahTmp = Convert.ToDouble(FahStr);
```
iShauny, line 287: Variables should be likeThis not LikeThis ^^

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
...
+            Regex miReg = new Regex("\\b(\\d+)(mi)\\b", RegexOptions.IgnoreCase);
+            Regex kmReg = new Regex("\\b(\\d+)(km)\\b", RegexOptions.IgnoreCase);
+            Regex kgReg = new Regex("\\b(\\d+)(kg)\\b", RegexOptions.IgnoreCase);
+            Regex lbReg = new Regex("\\b(\\d+)(lbs)\\b", RegexOptions.IgnoreCase);
+
+            if (fahReg.Match(input).Success)
+            {
+                Match m = fahReg.Match(input);
+
+                Group g = m.Groups[1];
```
iShauny, line 284: is there a need to use groups? I believe match returns the first result anyway, or am I missing something?

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
...
+
+                await Context.Channel.SendMessageAsync("", false, builder.Build());
+            }
+            else if (celReg.Match(input).Success)
+            {
+                Match m = celReg.Match(input);
+
+                Group g = m.Groups[1];
+
+                string celStr = Convert.ToString(g);
```
iShauny, line 307: where are these variables used? I can't see

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
...
+                double celTmp = Convert.ToDouble(celStr);
+
+                Temperature Cel = Temperature.FromDegreesCelsius(celTmp);
+                double Fah = Cel.DegreesFahrenheit;
+
+                EmbedBuilder builder = new EmbedBuilder()
+                {
+                    Title = "Temperature conversion",
+                    Description = $"🌡 {(Temperature)Cel} is equal to {(double)Math.Round(Fah, 2, MidpointRounding.ToEven)}°F.",
```
iShauny, line 316: I dont think you need explicit casting here as they are already explicitly defined during initialisation.

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
...
+                    Color = EMBED_COLOR
+                };
+
+                await Context.Channel.SendMessageAsync("", false, builder.Build());
+            }
+            else if (miReg.Match(input).Success)
```
iShauny, line 322: (optional) You can also use .IsMatch instead of .Success

Floofbot/Modules/Utilities.cs
```diff
@@ -256,4 +258,163 @@ public async Task ServerList()
...
+                string miStr = Convert.ToString(g);
+                double miTmp = Convert.ToDouble(miStr);
+
+                Length mi = Length.FromMiles(miTmp);
+                double km = mi.Kilometers;
+
+                EmbedBuilder builder = new EmbedBuilder()
+                {
+                    Title = "Length conversion",
+                    Description = $"📏 {(Length)mi} is equal to {(double)Math.Round(km, 3, MidpointRounding.ToEven)}Km.",
```
iShauny, line 337: Actually, looking at it now, I would just move the rounding to the variable declaration

`double km = Math.Round(mi.Kilometers, 3, MidpointRounding.ToEven)`

iShauny (changes requested):
I think this looks a lot better! A few picky things from me that I think would help keep formatting consistent. We wanna try to make sure the formatting through the whole program is the same! Good work though! I would like to see this reworked in a future PR though to remove the if statements for a tidier solution to allow expandability for other units

Floofbot/Modules/Utilities.cs
Shelme74: Actually, the match returns the whole string, that's why I needed to use the groups

Floofbot/Modules/Utilities.cs
Shelme74: the `g` var is used in the `Convert.ToString()` and the `celStr` is used in the next line in the `ToDouble` conversion. Then it's used to convert units.

I know, that's not optimal ><

Floofbot/Modules/Utilities.cs
iShauny: You can use positive lookahead instead of groups, eg

`\d+(?=f )`

Floofbot/Modules/Utilities.cs
Shelme74: Actually it is, because it's a class. However, related to the above comment, I'll make it a command to it won't be needed anymore.

Shelme74:
Code has been updated to take comments into account. I think it still can be optimized, but it works fine there.

Shelme74:
Changed have been made to improve the code, including the addition of negative temperatures support.

Ready for review

Floofbot/Modules/Utilities.cs
```diff
@@ -255,5 +257,127 @@ public async Task ServerList()
                 Trash = true
             }, true);
         }
+        [Command("convert")]
+        [Alias("conv")]
+        [Summary("Converts units to other units, such as Celcius to Fahrenheit.")]
+
+        public async Task convert(string input)
```
iShauny, line 264: Remember a parameter description and also use [Remainder] to take the whole rest of the command, refer to other commands in utilities for this

Floofbot/Modules/Utilities.cs
```diff
@@ -255,5 +257,127 @@ public async Task ServerList()
...
+            Regex lbReg = new Regex(@"\d+(?=lbs)", RegexOptions.IgnoreCase);
+
+            if (fahReg.IsMatch(input))
+            {
+                double fahTmp = Convert.ToDouble(Regex.Match(input, @"-?\d+").Value);
+
+                Temperature fah = Temperature.FromDegreesFahrenheit(fahTmp);
+                double cel = Math.Round(fah.DegreesCelsius, 2, MidpointRounding.ToEven);
+
+                EmbedBuilder builder = new EmbedBuilder()
```
iShauny, line 280: [OPTIONAL]

This is good for a first iteration, however, might be good to take the embed builder separate, then append to the description, then remove the 'else' in the 'else if' so that multiple conversations can be made on the same command. eg 

At the top of the command
```
string embedDescription = "";
```

Then in each command:

```
EmbedDescription += "🌡 {fah} is equal to {cel}°C.\n";
```

then at the end after all the ifs 
```
                EmbedBuilder builder = new EmbedBuilder()
                {
                    Title = "Conversion",
                    Description = embedDescription,
                    Color = EMBED_COLOR
                };
```
I think this is a cleaner, more handy way to use this command

iShauny (changes requested):
I think just do the parameter description change (refer to other commands with parameters for this) and maybe put the [Remainder] part and then it will be good to go, would be great if you could implement the other change but its optional for a first iteration

Floofbot/Modules/Utilities.cs
Shelme74: I updated the code with the new variable, however I didn't remove the `else` because the code doesn't handle more than one parameter at this time. I shall work on it later, though.

Shelme74: > however I didn't remove the `else` because the code doesn't handle more than one parameter at this time. I shall work on it later, though.
 Don't mind this, it works with the `[Remainder]`

Shelme74: It works but it keeps only the first number inputted, I'll work on it later on to see where it comes from.
Probably something to do with the Regex'ex.

Shelme74:
Might be good to go, let me know if other changes are required.

Shelme74:
Let me know if any changes are required so I'll work on them, or if something is missing

Shelme74:
Well, is there any news here? PR is, I think, ready to be merged since last August, and as I don't have write access I can't do it myself...
If something still needs to be updated, I'll take time to do so.

iShauny:
@Shelme74 resolve the merge conflict and I'll merge

Shelme74:
Conflicts have been fixed, I took the opportunity to fix a missing bracket which was probably accidentally deleted while removing conflict markers. Also UnitsNET has been updated to its latest version.

Llusus:
Merging.

PR #139, opened 2020-12-10
Moved message sending to end of command

iShauny:
This will allow the ban to be processed even if the bot cannot send messages to the user or channel

### DIFF
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -12,6 +12,7 @@ using Discord.Addons.Interactive;
 using Microsoft.EntityFrameworkCore;
 using Serilog;
 using System.Drawing.Printing;
+using Discord.Net;
 
 namespace Floofbot.Modules
 {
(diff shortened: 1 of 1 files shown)


Floofbot/Modules/Administration.cs
```diff
@@ -84,6 +78,13 @@ public class Administration : InteractiveBase
             builder.AddField("Moderator", $"{Context.User.Username}#{Context.User.Discriminator}");
 
             await Context.Channel.SendMessageAsync("", false, builder.Build());
+
+            //sends message to user
+            builder.Title = "⚖️ Ban Notification";
+            builder.Description = $"You have been banned from {Context.Guild.Name}";
+            builder.AddField("Reason", reason);
+            builder.Color = ADMIN_COLOR;
+            await badUser.SendMessageAsync("", false, builder.Build());
```
jkchen2, line 87: This embed would still have the fields added above (`User ID` and `Moderator`), no?

For safety, I'd suggest using `new EmbedBuilder` again for the ban embed.

jkchen2 (changes requested):

Floofbot/Modules/Administration.cs
```diff
@@ -65,25 +67,33 @@ public class Administration : InteractiveBase
...
-            //sends message to user
-            EmbedBuilder builder = new EmbedBuilder();
-            builder.Title = "⚖️ Ban Notification";
-            builder.Description = $"You have been banned from {Context.Guild.Name}";
-            builder.AddField("Reason", reason);
-            builder.Color = ADMIN_COLOR;
-            await badUser.SendMessageAsync("", false, builder.Build());
+            try
+            {
+                //sends message to user
```
jkchen2, line 72: Best practice is typically to limit the try/catch block to only the section of code that raises the exception. But I'll let this slide

Floofbot/Modules/Administration.cs
```diff
@@ -128,25 +138,32 @@ public class Administration : InteractiveBase
...
             }
 
-            //sends message to user
-            EmbedBuilder builder = new EmbedBuilder();
-            builder.Title = "⚖️ Ban Notification";
-            builder.Description = $"You have been banned from {Context.Guild.Name}";
-            builder.AddField("Reason", reason);
-            builder.Color = ADMIN_COLOR;
-            await badUser.SendMessageAsync("", false, builder.Build());
+            try
```
jkchen2, line 141: Lot of duplicate code. Future want: put this into a helper function

jkchen2 (approved):

PR #141, opened 2021-01-01
Moved rules gate to an event handler

iShauny:
Rules gate is no longer a command and now only requires a user to respond with "yes" in there message to accept the rules

### DIFF
--- a/Floofbot/Services/EventHandlerService.cs
+++ b/Floofbot/Services/EventHandlerService.cs
@@ -23,6 +23,9 @@ namespace Floofbot.Services
         // list of announcement channels
         private List<ulong> announcementChannels;
 
+        // rules gate config
+        private Dictionary <string, ulong> rulesGateConfig = BotConfigFactory.Config.RulesGate;
+
 
         public EventHandlerService(DiscordSocketClient client)
         {
@@ -124,8 +127,8 @@ namespace Floofbot.Services
                 return;
 
             // rules gate info
-            ulong rulesChannelId = 225980129799700481;
-            ulong readRulesRoleId = 494149550622375936;
+            ulong rulesChannelId = rulesGateConfig["RulesChannel"];
+            ulong readRulesRoleId = rulesGateConfig["RulesRole"];
             string rulesBypassString = "yes";
 
             if (msg.Channel.Id == rulesChannelId && userMsg.Content.ToLower().Contains(rulesBypassString)) 
--- a/Floofbot/Services/EventHandlerService.cs
+++ b/Floofbot/Services/EventHandlerService.cs
@@ -42,6 +42,7 @@ namespace Floofbot.Services
             _client.GuildMemberUpdated += GuildMemberUpdated;
             _client.UserUpdated += UserUpdated;
             _client.MessageReceived += OnMessage;
+            _client.MessageReceived += RulesGate; // rfurry rules gate
             _client.ReactionAdded += _nicknameAlertService.OnReactionAdded;
 
             // a list of announcement channels for auto publishing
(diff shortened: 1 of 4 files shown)


Floofbot/Services/EventHandlerService.cs
```diff
@@ -114,6 +115,28 @@ private async Task CheckUserAutoban(IGuildUser user)
...
+
+        // rfurry rules gate
+        public async Task RulesGate(SocketMessage msg)
+        {
+            var userMsg = msg as SocketUserMessage;
+            if (msg == null || msg.Author.IsBot)
+                return;
+
+            // rules gate info
+            ulong rulesChannelId = 225980129799700481;
```
jkchen2, line 127: All 4 of these values should be in the config

(See below, you might be able to get rid of `rulesServerId`)

Floofbot/Services/EventHandlerService.cs
```diff
@@ -114,6 +115,28 @@ private async Task CheckUserAutoban(IGuildUser user)
...
+            ulong rulesServerId = 225980129799700481;
+            ulong readRulesRoleId = 494149550622375936;
+            string rulesBypassString = "yes";
+
+            if (msg.Channel.Id == rulesChannelId && userMsg.Content.ToLower().Contains(rulesBypassString)) 
+            {
+                var user = (IGuildUser)msg.Author;
+                SocketGuild guild = _client.GetGuild(rulesServerId);
```
jkchen2, line 135: Is using `GetGuild` necessary? Can you use `user.Guild` (and cast to `SocketGuild` if necessary) instead? The assumption is that if the channel ID matches the defined rules channel, the message already originates from the target server.

Floofbot/Services/EventHandlerService.cs
```diff
@@ -114,6 +115,28 @@ private async Task CheckUserAutoban(IGuildUser user)
...
+            if (msg == null || msg.Author.IsBot)
+                return;
+
+            // rules gate info
+            ulong rulesChannelId = 225980129799700481;
+            ulong rulesServerId = 225980129799700481;
+            ulong readRulesRoleId = 494149550622375936;
+            string rulesBypassString = "yes";
+
+            if (msg.Channel.Id == rulesChannelId && userMsg.Content.ToLower().Contains(rulesBypassString))
```
jkchen2, line 132: What's the rationale behind using `Contains`? Why not use an exact match?

jkchen2 (changes requested):

Floofbot/Services/EventHandlerService.cs
iShauny: I was unsure about adding it to the config, as its a pretty specific thing for a specific server

Floofbot/Services/EventHandlerService.cs
iShauny: changed

Floofbot/Services/EventHandlerService.cs
iShauny: The largest issue we are having with the rules gate is due to typos or people writing it in different ways "eg ihavereadtherules"

If we change it to yes, its much easier to do, however, it would also be good if it caught cases such as "yes I agree" or "yes I do" etc

Floofbot/Services/EventHandlerService.cs
jkchen2: It's a good practice thing. Not having stuff like this configurable has bitten us in the collective ass previously, and I'd like to avoid something like that happening even if it seems very unlikely (which it is).

Floofbot/Services/EventHandlerService.cs
jkchen2: Ah, alright. That's a fair way of putting it.

jkchen2 (approved):

PR #146, opened 2021-01-18
Created Welcome Gate Service

iShauny:
This auto adds a predefined role to a user when they pass the welcome gate

### DIFF
--- a/Floofbot/Services/EventHandlerService.cs
+++ b/Floofbot/Services/EventHandlerService.cs
@@ -43,6 +43,7 @@ namespace Floofbot.Services
             _client.UserJoined += UserJoined;
             _client.UserLeft += UserLeft;
             _client.GuildMemberUpdated += GuildMemberUpdated;
+            _client.GuildMemberUpdated += HandleWelcomeGate; // welcome gate handler
             _client.UserUpdated += UserUpdated;
             _client.MessageReceived += OnMessage;
             _client.MessageReceived += RulesGate; // rfurry rules gate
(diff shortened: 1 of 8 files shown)


Floofbot/Services/EventHandlerService.cs
```diff
@@ -691,6 +692,34 @@ public Task GuildMemberUpdated(SocketGuildUser before, SocketGuildUser after)
...
+                var guild = after.Guild;
+                WelcomeGate serverConfig = floofDb.WelcomeGateConfigs.Find(guild.Id);
+
+                if (serverConfig == null || serverConfig.Toggle == false || serverConfig.RoleId == null) // disabled
+                    return;
+
+                try
+                {
+                    var userRole = guild.GetRole((ulong)serverConfig.RoleId);
+                    if (userRole == null)
```
jkchen2, line 711: May be worth adding an extra debug log here so that it's not a silent fail. Should help with debugging problems with role assignment.

Floofbot/Services/WelcomeGateService.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Discord.WebSocket;
```
jkchen2, line 1: Why the separate file? Looks like all the logic is handled in `EventHandlerService`

jkchen2 (changes requested):

Floofbot/Services/WelcomeGateService.cs
iShauny: image

Floofbot/Services/EventHandlerService.cs
```diff
@@ -43,7 +45,7 @@ public EventHandlerService(DiscordSocketClient client)
             _client.UserJoined += UserJoined;
             _client.UserLeft += UserLeft;
             _client.GuildMemberUpdated += GuildMemberUpdated;
-            _client.GuildMemberUpdated += HandleWelcomeGate; // welcome gate handler
+            _client.GuildMemberUpdated += _welcomeGateService.HandleWelcomeGate; ; // welcome gate handler
```
jkchen2, line 48: Double semicolon

jkchen2 (approved):

PR #147, opened 2021-01-18
Added User Role Retention

iShauny:
- Roles will be logged when a user leaves a server and returns on rejoining
- This prevents users from leaving to avoid mutes
- Note: the bot cannot return roles that it is not permitted to give (eg higher than its highest roles)


Floofbot/Services/UserRoleRetentionService.cs
```diff
@@ -0,0 +1,95 @@
...
+            var oldUserRoles = GetUserRoles(user);
+
+            if (oldUserRoles != null) // user actually had old roles
+            {
+                var oldUserRolesList = oldUserRoles.Split(",");
+
+                // restore each role 1 by 1
+                foreach (string roleId in oldUserRolesList)
+                {
+                    SocketRole role = GetRole(roleId, user.Guild);
```
jkchen2, line 75: Can you use the built-in [GetRole](https://docs.stillu.cc/api/Discord.IGuild.html#Discord_IGuild_GetRole_System_UInt64_) instead?

Floofbot/Services/UserRoleRetentionService.cs
```diff
@@ -0,0 +1,95 @@
...
+                    if (role == null) // role does not exist
+                        continue;
+                    else
+                    {
+                        try
+                        {
+                            await user.AddRoleAsync(role);
+                        }
+                        catch
+                        {
```
jkchen2, line 86: Again, to avoid silent failure, I'd suggest logging this to help with debugging. Same with L77

Floofbot/Services/UserRoleRetentionService.cs
```diff
@@ -0,0 +1,95 @@
...
+        public async Task RestoreUserRoles(IGuildUser user)
+        {
+            var oldUserRoles = GetUserRoles(user);
+
+            if (oldUserRoles != null) // user actually had old roles
+            {
+                var oldUserRolesList = oldUserRoles.Split(",");
+
+                // restore each role 1 by 1
+                foreach (string roleId in oldUserRolesList)
```
jkchen2, line 73: I wonder if there should be a check for roles that _shouldn't_ be reassigned. For example, users that leave and join an excessive amount of times usually end up having to wait a week before they can be re-verified.

Reassigning roles makes sense for stuff like the muted and user roles (and maybe the moderator role), but other roles should probably be left alone IMO.

jkchen2 (changes requested):

Floofbot/Services/UserRoleRetentionService.cs
```diff
@@ -72,7 +66,7 @@ public async Task RestoreUserRoles(IGuildUser user)
                 // restore each role 1 by 1
                 foreach (string roleId in oldUserRolesList)
                 {
-                    SocketRole role = GetRole(roleId, user.Guild);
+                    IRole role = user.Guild.GetRole(Convert.ToUInt64(roleId));
 
                     if (role == null) // role does not exist
```
jkchen2, line 71: Sorry, probably should've left a comment on this line but I was also requesting a debug log here as well.

jkchen2 (approved):

PR #149, opened 2021-01-20
Fix error on trying to reassign the @everyone role

jkchen2:


iShauny (approved):
Could remove brackets if you want

jkchen2:
> Could remove brackets if you want

For the sake of consistency, I opted to use brackets. Imagine how weird it would look to mix brackets in an `if` `else if` chain

PR #152, opened 2021-02-27
Nickname alert now shows detected filtered words

iShauny:
- Nickname alert now shows the filtered words it detected in a users name/nickname
- Made nickname alert service independent from the event handler toggling and logging

### DIFF
--- a/Floofbot/Services/WordFilterService.cs
+++ b/Floofbot/Services/WordFilterService.cs
@@ -13,13 +13,13 @@ namespace Floofbot.Services
         List<FilteredWord> _filteredWords;
         DateTime _lastRefreshedTime;
 
-        public bool hasFilteredWord(FloofDataContext floofDb, string messageContent, ulong serverId) // names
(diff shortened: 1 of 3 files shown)


Floofbot/Services/WordFilterService.cs
```diff
@@ -30,16 +30,21 @@ class WordFilterService
...
                     RegexOptions.IgnoreCase | RegexOptions.Singleline);
                 if (r.IsMatch(messageContent))
                 {
-                    return true;
+                    DetectedWords.Add(filteredWord.Word);
                 }
             }
-            return false;
+            if (DetectedWords.Count() == 0)
+                return null;
```
jkchen2, line 45: While this would work, I'd say it's generally more favorable to return an empty list as it's conceptually more correct (i.e. the number of filtered words is zero).

jkchen2 (approved):
Just as a general comment, it would be nice to extend this functionality to deleted user messages as well.

PR #155, opened 2021-06-12
Added silent kick command

iShauny:
This command will kick a user from the server without notifying them


jkchen2 (approved):
Looks good to me, though I'm not sure how useful this command will be. How is it functionally different from just kicking a user the normal way?

PR #156, opened 2021-06-27
Fixed nickname alert regex parsing

iShauny:
Replaced the nickname alert detection to String.Contains() to prevent filtered words containing regex from being parsed

### DIFF
--- a/Floofbot/Services/WordFilterService.cs
+++ b/Floofbot/Services/WordFilterService.cs
@@ -34,9 +34,7 @@ namespace Floofbot.Services
 
             foreach (var filteredWord in _filteredWords)
             {
-                Regex r = new Regex($"{filteredWord.Word}",
-                    RegexOptions.IgnoreCase | RegexOptions.Singleline);
-                if (r.IsMatch(messageContent))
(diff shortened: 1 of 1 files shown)


Floofbot/Services/WordFilterService.cs
```diff
@@ -74,7 +72,7 @@ class WordFilterService
             {
                 Regex r = new Regex(@$"\b({filteredWord.Word})\b",
                     RegexOptions.IgnoreCase | RegexOptions.Singleline);
-                if (r.IsMatch(messageContent))
+                if (r.IsMatch(Regex.Escape(messageContent)))
```
jkchen2, line 75: If I'm reading this right, you're escaping regex on the message content, not on the filtered word. I assume it should be `Regex.Escape(filteredWord.Word)` instead?

jkchen2 (changes requested):

Floofbot/Services/WordFilterService.cs
iShauny: yes

iShauny: Resolved

jkchen2 (approved):

PR #161, opened 2021-06-28
Fix unclosed quote

jkchen2:


jkchen2:
Whoops, I meant unclosed parenthesis. Oh well, you get the idea

PR #162, opened 2021-09-11
Fixed some titles on the help page + added about command with github link!

lunarlynix:
I noticed some of the help pages were not having correct naming and instead only had the group module name. I added the Name Tag for the ones that did not have one!

### DIFF
--- a/Floofbot/Floofbot.csproj
+++ b/Floofbot/Floofbot.csproj
@@ -12,10 +12,10 @@
   <ItemGroup>
     <FrameworkReference Include="Microsoft.AspNetCore.App" />
     <ProjectReference Include="../Discord.Addons.Interactive/Discord.Addons.Interactive.csproj" />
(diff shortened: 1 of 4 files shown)


iShauny:
Some investigation will need to be done to ensure library updates do not cause issues before we can update the libraries

lunarlynix:
We can check together if ya want.

jkchen2:
I don't think Kuro is around to approve this right now, but it seems to have the green light from the mod team.

A significant amount of work will likely be required to tackle #163, and while I like the idea of using slash commands for better help integration, the risk of potentially breaking some core functionality over better help integration doesn't seem worth it at the moment. Plus, we'd have to do some more investigations into the complexity of this task, as well as whether or not Discord.Net will offer a stable API before we attempt an implementation. That said, slash commands is definitely the future (present?) of interacting with bots, so we should get on top of this before next year.

PR #164, opened 2021-12-13
Update UriKind to Relative for birb command

the-og-gear:
The birb command has been spitting out that it is unavailable for quite a while now. Doing a little digging, I found that the file key provided by the API is merely a file, not a full URI. This caused `UriKind.Absolute` to fail the check in `IsWellFormedString`.

This update adjusts to relative so as to not change any logic going on while allowing responses to be provided once again.
image


the-og-gear:
Additional note, I could've built the URI before it got passed to the check but decided against doing this. There's not really a benefit when it's going to be built in the first place, and we can manually verify it to be a valid URI at that point (it's hard coded, after all).

jkchen2 (approved):
I don't normally do the PRs but this is literally a single line change

PR #165, opened 2022-01-07
Finally upgrade dependencies + Fix style warnings

the-og-gear:
Floofbot has been running on a wildly old version of Discord.Net for a good while now. Let's fix that! 
I've also went through and updated what was required in Discord.Addons.Interactive (including the license field, which wasn't properly set), solved the issues that caused in the primary project, and while I was at it I updated styles and provided suppressions for the present issues to ensure code analysis doesn't give messages.

All of this because I intended to add a command to pull guild avatars. Huh.

The preparation of updating dependencies was required for the guild avatars to be feasible in the first place, which prompted this change.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+﻿[*.cs]
+
+# IDE0008: Use explicit type
+csharp_style_var_when_type_is_apparent = true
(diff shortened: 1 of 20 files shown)


the-og-gear:
Ironically, the build fails because to update the license field of the project you need to manually remove the `licenseUrl` field from the given file. Visual Studio no longer offers a way to do this cleanly and it's not part of the repository. Beyond this, it builds fine. I'll see what I can do to ensure this failing build doesn't happen but it might not be much.

Also, this will absolutely, 100% cause merge conflicts somewhere with other PRs. Let me know if you determine this PR is to be merged after those and I'll push another couple commits to amend that.

the-og-gear:
You know what, let's see if I can get away with not including this licensing thing. Closing for some addtl research to be done.

PR #169, opened 2023-12-17
Update dependencies and modules

iShauny:
Update to use updated dependencies and code refactor to allow modules to log events. User time in server removed from leaving event to accommodate discord API changes with caching. Future feature will add a user cache service that runs regularly to cache user data.

### DIFF
--- a/Floofbot/Services/EventHandlerService.cs
+++ b/Floofbot/Services/EventHandlerService.cs
@@ -253,8 +253,9 @@ namespace Floofbot.Services
             });
             return Task.CompletedTask;
         }
-        public Task MessageDeleted(Cacheable<IMessage, ulong> before, ISocketMessageChannel chan)
+        public Task MessageDeleted(Cacheable<IMessage, ulong> before, Cacheable<IMessageChannel, ulong>  channel)
         {
+            IMessageChannel chan = channel.Value;
             var _ = Task.Run(async () =>
             {
                 try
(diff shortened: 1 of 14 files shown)


Floofbot/Services/EventHandlerService.cs
```diff
@@ -491,27 +496,18 @@ public Task UserLeft(IGuildUser user)
...
-                    Discord.ITextChannel channel = await GetChannel(user.Guild, "UserLeftChannel");
+                    Discord.ITextChannel channel = await GetChannel(guild, "UserLeftChannel");
                     if (channel == null)
                         return;
 
                     var embed = new EmbedBuilder();
                     embed.WithTitle($"❌ User Left | {user.Username}#{user.Discriminator}")
                          .WithColor(Color.Red)
                          .WithDescription($"{user.Mention} | ``{user.Id}``");
-                    if (user.JoinedAt != null)
```
jkchen2, line 507: So what happened to `JoinedAt`? I can't seem to find older docs on previous versions of Discord.NET, and there is no entry for `JoinedAt` for version 3.

jkchen2 (approved):
I don't have the means to test these changes, so I'll have to assume you've done the proper testing on your end. But as far as I can tell, these changes look okay to me.

Floofbot/Services/EventHandlerService.cs
iShauny: Because it's changed from guilduser to socketuser, this no longer knows the join date as the user isn't associated with a guild, couldn't think of alternative solution but to remove it

the-og-gear:
Rejoice, for it has updated

Floofbot/Services/EventHandlerService.cs
jkchen2: This is really late, but you can actually get the `SocketGuildUser` by using [`GetUser`](https://discordnet.dev/api/Discord.WebSocket.SocketGuild.html#Discord_WebSocket_SocketGuild_GetUser_System_UInt64_).

PR #170, opened 2023-12-18
Various bug fixes and feature requests over the last couple years

iShauny:
Addresses various bug fixes noted by users and moderators:

CHANGES:
- Ensure a username/id is always resolved for all admin commands
- Add protection against a moderator performing actions on themselves or the bot
- Added a check for pruning messages that the message is not a null object (incase this message is deleted whilst the bot is purging or exists in cache but not in discord)
- Updated filter message warn to be less aggressive
- Added try/catch to catch errors where response is not as expected for parsing or permissions are missing
- Added unban command
- Add command aliases to the help page
- Added a moderation command to fetch a user id
- Added event handlers to make the bot join all threads that are updated or created (avoided running a service to check the bot is in all threads to avoid API rate limiting)
- Enforce SFW FA links
- Set the number of days to prune for pb command to 7

(PR https://github.com/bealsbe/Floofbot/pull/170 needs to be accepted first before this PR)

### DIFF
--- a/Floofbot/Modules/Administration.cs
+++ b/Floofbot/Modules/Administration.cs
@@ -13,6 +13,8 @@ using Microsoft.EntityFrameworkCore;
 using Serilog;
 using System.Drawing.Printing;
 using Discord.Net;
+using System.Security.Cryptography;
+using Microsoft.Extensions.Configuration.UserSecrets;
 
 namespace Floofbot.Modules
 {
@@ -67,6 +69,17 @@ namespace Floofbot.Modules
                 }
             }
 
+            if (badUser.Id == Context.Client.CurrentUser.Id)
+            {
+                await Context.Channel.SendMessageAsync("⚠️ I cannot ban myself.");
+                return;
+            }
+            if (badUser.Id == Context.User.Id)
+            {
+                await Context.Channel.SendMessageAsync("⚠️ For safety reasons, I cannot allow you to ban yourself.");
+                return;
+            }
+
             try
             {
                 //sends message to user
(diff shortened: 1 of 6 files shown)


Floofbot/Modules/Administration.cs
```diff
@@ -82,17 +95,81 @@ public class Administration : InteractiveBase
...
+                EmbedBuilder modEmbedBuilder = new EmbedBuilder();
+                modEmbedBuilder = new EmbedBuilder();
+                modEmbedBuilder.Title = (":shield: User Banned");
+                modEmbedBuilder.Color = ADMIN_COLOR;
+                modEmbedBuilder.Description = $"{badUser.Username}#{badUser.Discriminator} has been banned from {Context.Guild.Name}";
+                modEmbedBuilder.AddField("User ID", badUser.Id);
+                modEmbedBuilder.AddField("Moderator", $"{Context.User.Username}#{Context.User.Discriminator}");
+                await Context.Channel.SendMessageAsync("", false, modEmbedBuilder.Build());
+            }
+            catch
```
jkchen2, line 112: This goes for all the try/catch blocks in this PR, but the try/catch should be centered around just the `AddBanAsync` call (to avoid accidentally catching errors from `SendMessageAsync`).

Additionally, the `catch` should catch specific errors if possible, as otherwise the error information is lost. [A generic catch-all `catch` should at the very least have some way to print the serialized error type and message](https://stackoverflow.com/a/12911864).

I don't know if there's a good way to see which exceptions each function can throw (docs don't seem to be much help). [`HttpException`](https://discordnet.dev/api/Discord.Net.HttpException.html) might be helpful to intercept permissions errors, but I'd suggest looking through the docs to see what else makes sense to catch ([`RateLimitedException`](https://discordnet.dev/api/Discord.Net.RateLimitedException.html) might be another good one).

Floofbot/Modules/Help.cs
```diff
@@ -113,10 +115,10 @@ public async Task HelpCommand([Summary("module name")] string requestedModule)
                     }
                     pages.Add(new PaginatedMessage.Page
                     {
-                        Author = new EmbedAuthorBuilder { Name = cmd.Name },
+                        Title = "Command Aliases: " + String.Join(", ", cmd.Aliases.ToList()),
                         Fields = new List<EmbedFieldBuilder>(fields),
                         Description = cmd.Summary ?? "No command description available"
-                    });
+                    });
```
jkchen2, line 121: Trailing space

jkchen2 (changes requested):
Like with the last PR, I don't have an easy means to test these changes. I trust that you've done the necessary testing to verify it behaves as expected.

Floofbot/Modules/Help.cs
iShauny: Trailing space removed

Floofbot/Modules/Administration.cs
iShauny: I don't see exact specific types of exceptions. I considered adding those two aforementioned exceptions - but thought it more fitting to add a generic exception handler and updated it to spit out the exception, so it will at least catch all exceptions and spit it out. Since these are admin commands anyway - a bot admin will see the details in the appropriate mod chat anyway.

iShauny: let me know if I missed any of the exception handlers, I think I caught them all and updated, and whether you agree/disagree with my thoughts ^^

iShauny: Also added additional exception handlers to the other fun commands since some were missing

jkchen2 (approved):

